PR #4, opened 2017-06-15
try to determine region automatically if not set:

dfuentes:
- Will attempt to use the ec2 metadata API to look up current region
  if AWS_REGION environment variable is not set


achille-roussel:
❤️

PR #8, opened 2017-07-07
Allow multiple service names to be specified

ejcx:


cmd/exec.go
```diff
@@ -42,27 +42,26 @@ func execRun(cmd *cobra.Command, args []string) error {
 		return errors.New("must specify command to run")
 	}
 
-	service := strings.ToLower(args[0])
```
Pryz: Why are we getting rid of the lowering here ?

cmd/exec.go
```diff
@@ -42,27 +42,26 @@ func execRun(cmd *cobra.Command, args []string) error {
...
+		if err := validateService(service); err != nil {
+			return err
+		}
+		secrets, err := secretStore.List(service, true)
+		if err != nil {
+			return err
+		}
+		for _, secret := range secrets {
+			env.Set(strings.ToUpper(key(secret.Meta.Key)), *secret.Value)
+		}
```
Pryz: We should not probably fail everything if only 1 secret is not accessible.

ejcx:
I believe chamber is already sneak silent

```
chamber exec does-not-exist
```

Will happily execute your program and not log any issues

cmd/exec.go
ejcx: Whoops. I missed this one.

ejcx: That's a really good catch, btw.

cmd/exec.go
abraithwaite: I'm not sure.  I'd expect a failure here.

Worst case scenario is you have two variables, one which overrides the other when set but the override doesn't get set properly.  You continue happily with the original, unaware that anything's gone wrong.  If you don't fail everything that it will probably get discovered in production and via logs instead of failing to start the first time.

cmd/exec.go
```diff
@@ -42,27 +42,27 @@ func execRun(cmd *cobra.Command, args []string) error {
...
 	}
 
+	env := environ(os.Environ())
 	secretStore := store.NewSSMStore()
-	secrets, err := secretStore.List(service, true)
-	if err != nil {
-		return err
-	}
+	for _, service := range args {
+		if err := validateService(service); err != nil {
```
abraithwaite: The logic for validateService changed when you moved the toLower below this function.

abraithwaite: Evan confirmed offline that this was checked. 👍

ejcx:
Alright! I will merge this.

PR #9, opened 2017-07-10
Provide a noop environment variable to bypass chamber

ejcx:
Chamber users are clamboring for this feature. They really want to be able to
painless run chamber in development without needing to pass Env variables from
aws-vault to their container, and without needing any special bash-fu to make
things work.

Provide CHAMBER_NOOP to just execute whatever command is passed to chamber.

I think one improvement could be logging loudly if chamber is being run with
this flag set. This would make debugging production issues easier in case
CHAMBER_NOOP has been set in production.

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -12,6 +12,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	NoopEnv = "CHAMBER_NOOP"
+)
+
 var (
 	ErrCommandMissing = errors.New("must specify command to run")
 )
@@ -25,8 +29,8 @@ var execCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(execCmd)
-
 }
+
 func execRun(cmd *cobra.Command, args []string) error {
 	dashIx := cmd.ArgsLenAtDash()
 	if dashIx == -1 {
@@ -54,10 +58,17 @@ func execRun(cmd *cobra.Command, args []string) error {
 		commandArgs = commandPart[1:]
 	}
 
-	secretStore := store.NewSSMStore()
-	secrets, err := secretStore.List(service, true)
-	if err != nil {
-		return err
+	// If NoopEnv is set then we do not want to actually fetch and list our
+	// secrets from AWS. This is so development services have an easy escape
+	// hatch to turn off chamber from fetching secrets.
+	var secrets []store.Secret
+	if _, ok := os.LookupEnv(NoopEnv); !ok {
+		secretStore := store.NewSSMStore()
+		secretList, err := secretStore.List(service, true)
+		if err != nil {
+			return err
+		}
+		secrets = secretList
 	}
 
 	env := environ(os.Environ())


ejcx:
Sorry @dfuentes. I'm sure this is not a feature you are excited about, but running chamber locally has consistently been a pain point for people.

I made a small bash alias that passes aws-vault creds to docker so that chamber can authenticate as individuals, but people are not excited about it and really want chamber to just fix it

```
docker run \
  --env "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID" \
  --env "AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY" \
  --env "AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN" \
  --env "AWS_SECURITY_TOKEN=$AWS_SECURITY_TOKEN" \
  --env "AWS_VAULT=$AWS_VAULT" \
  --env "AWS_REGION=$AWS_REGION" \
  --env "AWS_DEFAULT_REGION=$AWS_DEFAULT_REGION" \
  $1
```

Curious what your thoughts are.

dfuentes:
Some issues I have with this approach:

1) Doesn't really solve the problem (as I understand it).  If an application requires a secret from chamber/parameter store to run, making chamber a no-op still doesn't solve the problem of applications needing that value.  The likely way a user would then solve the problem is making some local run script or make target that invokes their docker container with the proper env vars hard coded.

My guess is that we could standardize the local run process in a way that makes it easy for users to invoke chamber with the correct auth locally, making this a one line change that they can make without constantly updating a run script/make target.

2)  I would really like to avoid solutions that fork the runtime flow between "local" and "remote" runs of an application.  Having two different code/execution paths for these set ups make bugs more likely as things drift apart.

I think we should step back and really look at the problem and brainstorm some solutions, instead of jumping to this one.  We could do this in person, or maybe open a github issue and we can discuss possible solutions there.

ejcx:
I don't think we actually need this anymore. If you are here, I have a shell script called `adr` in my `$HOME/bin` directory that has the following code

```
#!/bin/zsh
docker run \
  --env "AWS_ACCESS_KEY_ID=$AWS_ACCESS_KEY_ID" \
  --env "AWS_SECRET_ACCESS_KEY=$AWS_SECRET_ACCESS_KEY" \
  --env "AWS_SESSION_TOKEN=$AWS_SESSION_TOKEN" \
  --env "AWS_SECURITY_TOKEN=$AWS_SECURITY_TOKEN" \
  --env "AWS_VAULT=$AWS_VAULT" \
  --env "AWS_REGION=$AWS_REGION" \
  --env "AWS_DEFAULT_REGION=$AWS_DEFAULT_REGION" \
  -it $1 /bin/sh
```

adr stands for authenticated docker run. I use it like this.

```
$ docker build .
.....
Created image d34db33f

$ aws-vault exec development -- adr d34db33f
```

FernandoMiguel:
when using aws-vault, consider using the flag --server
this starts a daemon and will make it possible to run chamber from anywhere

(still requires  `export AWS_REGION=` )

PR #10, opened 2017-07-13
Add debian package with fpm and packagecloud

Pryz:


circle.yml
```diff
@@ -40,3 +42,5 @@ deployment:
         --workdir /go/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
         --entrypoint make
         segment/golang:latest release
+      - fpm -s dir -t deb -n chamber -v $(CIRCLE_TAG) ./dist/chamber--linux-amd64=/usr/bin/chamber
+      - package_cloud push segment/chamber/ubunty/xenial chamber_$(CIRCLE_TAG)_amd64.deb
```
abraithwaite: unbunty

Pryz: fixed

Makefile
```diff
@@ -27,12 +27,16 @@ release: gh-release clean dist
...
 	rm -rf ./dist
+	rm -f chamber_*.deb
 
 dist:
 	mkdir dist
 	GOOS=darwin GOARCH=amd64 go build -o dist/chamber-$(version)-darwin-amd64
 	GOOS=linux GOARCH=amd64 go build -o dist/chamber-$(version)-linux-amd64
 
+deb: dist
+	fpm -s dir -t deb -n chamber -v $(version) ./dist/chamber--linux-amd64=/usr/bin/chamber
```
dfuentes: I think the path should be `./dist/chamber-$(CIRCLE_TAG)-linux-amd64`.

circle.yml
```diff
@@ -40,3 +42,5 @@ deployment:
         --workdir /go/src/github.com/${CIRCLE_PROJECT_USERNAME}/${CIRCLE_PROJECT_REPONAME}
         --entrypoint make
         segment/golang:latest release
+      - fpm -s dir -t deb -n chamber -v $(CIRCLE_TAG) ./dist/chamber--linux-amd64=/usr/bin/chamber
```
dfuentes: Instead of putting these two lines in the circle yaml directly, they should be in the make file.  Then that make target should be invoked in the `deployment` section of the circle yaml so that we only build/push packages on tagged commits.  The deployment section already runs one make target, you should be able to just add an additional make target to that invocation.

PR #11, opened 2017-07-17
Clarify conflicting secrets

dfuentes:
This adds a warning printed to stderr if you have conflicting secret names from loading secrets from multiple services.

It also clarifies this behavior in the readme for `exec`.  I'm not super happy with the copy there, so if you have suggestions for improving it let me know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,9 @@ The history command gives a historical view of a given secret. This view is usef
 $ chamber exec <service...> -- <your executable>
 ```
 
-The purpose of this command is to be used inside your service's docker container. Exec sets up environment variables populated with all the latest versions of secrets for the given service and environment.
+`exec` populates the environment with the secrets from the specified services and executes the given command.  Secret keys are converted to upper case (for example a secret with key `secret_key` will become `SECRET_KEY`).
+
+Secrets from services are loaded in the order specified in the command.  For example, if you do `chamber exec app apptwo -- ...` and both apps have a secret named `api_key`, the `api_key` from `apptwo` will be the one set in your environment.
 
 ## Releasing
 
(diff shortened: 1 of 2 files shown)


cmd/exec.go
```diff
@@ -61,6 +62,11 @@ func execRun(cmd *cobra.Command, args []string) error {
 			return err
 		}
 		for _, secret := range secrets {
+			envVarKey := strings.ToUpper(key(secret.Meta.Key))
+
+			if env.IsSet(envVarKey) {
+				fmt.Fprintf(os.Stderr, "warning: overwriting environment variable %s\n", envVarKey)
+			}
 			env.Set(strings.ToUpper(key(secret.Meta.Key)), *secret.Value)
```
tamarrow: we can use `envVarKey` here

README.md
```diff
@@ -54,7 +54,9 @@ The history command gives a historical view of a given secret. This view is usef
 $ chamber exec <service...> -- <your executable>
 ```
 
-The purpose of this command is to be used inside your service's docker container. Exec sets up environment variables populated with all the latest versions of secrets for the given service and environment.
+Exec executes the given command in an environment where all secrets from the specified services are set.  Secret keys are converted to upper case (for example a secret with key `secret_key` will become `SECRET_KEY`).
```
tamarrow: How about `exec populates the environment with the secrets from the specified services and executes the given command. Secret keys....`?

cmd/exec.go
dfuentes: oops, good catch 0e586bb

README.md
dfuentes: Sounds better thanks :)

PR #12, opened 2017-08-01
circle: add ECR credentials to pull images from in the future

tamarrow:


tamarrow:
You're right, didn't realize this was a public repo.

PR #14, opened 2017-08-12
disable cgo in distributed versions:

dfuentes:
- So this can run in alpine containers


abraithwaite:
What is/was it using cgo for?

dfuentes:
Some parts of the `net` package use cgo.

PR #17, opened 2017-08-19
Read subcommand: Read one secret at a time.

ejcx:


sc-chad:
Sounds great to me!

dfuentes:
Looks good, some additional flags that may be useful to add (although we could do it later):

`-v/--version <int>` to read a specific version of the secret
`-q/--quiet` Print only the secret, not the whole table/metadata

sc-chad:
Both great suggestions.

ejcx:
LGTM, what say you?

PR #23, opened 2017-08-25
use paths instead of '.' separating

dfuentes:
This is a rough minimum edit change for using paths instead of dot separators.  If we go forward with this we should also:

- Update Describe and Get calls to use the Path options instead of string prefixing
- Bump to v2 as this is a breaking API change, probably put a notice in the readme as well

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,12 @@ Chamber does not currently read the value of "AWS_DEFAULT_REGION". See
 [https://github.com/aws/aws-sdk-go#configuring-aws-region](https://github.com/aws/aws-sdk-go#configuring-aws-region)
 for more details.
 
+### Using Path Based Keys
+
+If you'd prefer to use path based keys (`/service/key`) instead of the default period separated keys (`service.key`), you
+can set the environment variable `CHAMBER_USE_PATHS` to 1.  This environment variable must be set when writing and reading keys.
+
+
 ## Releasing
 
 To cut a new release, just push a tag named `v<semver>` where `<semver>` is a
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -55,7 +55,14 @@ func list(cmd *cobra.Command, args []string) error {
 }
 
 func key(s string) string {
+	_, usePaths := os.LookupEnv("CHAMBER_USE_PATHS")
+	if usePaths {
+		tokens := strings.Split(s, "/")
+		secretKey := tokens[2]
+		return secretKey
+	}
+
 	tokens := strings.Split(s, ".")
-	secretKey := tokens[1]
+ 	secretKey := tokens[1]
 	return secretKey
 }
--- a/store/ssmstore.go
+++ b/store/ssmstore.go
@@ -19,13 +19,20 @@ const (
 	DefaultKeyID = "alias/parameter_store_key"
 )
 
-// validKeyFormat is the format that is expected for key names inside parameter store
+// validPathKeyFormat is the format that is expected for key names inside parameter store
+// when using paths
+var validPathKeyFormat = regexp.MustCompile(`^\/[A-Za-z0-9-_]+\/[A-Za-z0-9-_]+$`)
+
+
+// validKeyFormat is the format that is expected for key names inside parameter store when
+// not using paths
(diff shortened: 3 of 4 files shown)


ejcx:
LGTM

rafaelmagu:
Any ETA to get this into a release? It'd be immensely helpful.

bsideup:
Could you please support multiple levels?

Would be really nice to have something like `/dev/services/foo/my-super-key`

whisller:
Why not to make it as a parameter? Some systems (as in our case) rely on name convention. So `{service}.{parameter}`. We use chamber to write/read parameters in few places but other systems read it as well.

dfuentes:
Wanted to give an update here on the current situation.  We will likely be merging and releasing this update within the next day under a new major version (v2.0).  The hold up is that we need to migrate our secrets here to use the new scheme and we want to provide a script for others to do so as well.

As far as nested paths and configurable separators, we have no current plans to do those.

RyanBowlby-Reflektion:
@dfuentes Any status update on your migration?

lox:
The cli validation still requires underscores from what I can tell.

jleclanche:
@dfuentes Looking forward to this as well. I'm about to adopt Chamber for our prod environment and as it turns out this is a blocker :)

lox:
Any updates or anything we can do to assist @dfuentes?

Vektah:
I've built this PR and uploaded a release on a fork for anyone spinning up new infrastructure:

https://github.com/99designs/chamber/releases/tag/2.0.0-pre

dfuentes:
Sorry for the radio silence on this for so long.  We don't use paths at Segment, so prioritizing this work over other stuff has been difficult.  I have updated the PR to only use path-based keys when an environment variable (`CHAMBER_USE_PATHS`) is set.  This is the easiest way for us to maintain backwards compatibility, which we require.

egarbi:
I've just tried this for the very first time:
```
$ ./chamber write 'test/nosensitive/naboo' host 'db-naboo.stack.local'
chamber error: Failed to validate service: Failed to validate service name 'test/nosensitive/naboo'.  Only alphanumeric, dashes, and underscores are allowed for service names
```
@dfuentes What am I doing wrong?

Vektah:
./chamber write naboo host 'db-naboo.stack.local'

Will write a parameter store key of naboo/host.

I guess service names themselves could also have separators but it becomes more ambiguous on where to split. Would the Param key foo/bar/baz/crux always use the last Param segnent as the env var name? I guess that is how envdir would work.

dfuentes:
@egarbi we only support a single level heirarchy, so service names can not have `/` in them

egarbi:
@dfuentes @Vektah thanks for your inputs. 
One more, would it be possible to allow "." (dots) in key names one paths are set to true?
Example: chamber write users 'enrique.garbi' mysupersecret
Think is all my user in aws apply to that convention in names and I would like to use ${aws:username} variable on policies to allow users to change his own secret

klindklind:
I'm using the latest release 2.1.0 and I'm little bit confused how to use CHAMBER_USE_PATHS mentioned in the documentation. Does that mean I can somehow set parameter path for chamber cli commands?

What I'm trying to accomplish here is to manage parameters with multi-level hierarchy and give fine grained parameter read permissions for different applications.

For example

Create parameters like
/myapp/dev/parameter1 (devvalue1)
/myapp/dev/parameter2 (devvalue2)
/myapp/prod/parameter1 (prodvalue1)

Give IAM permission for dev environment applications to only read /myapp/dev/*

Is there someway to manage this kind of multi-level hierarchy parameters with Chamber?

egarbi:
@klindklind As far as I know, only multi-level hierarchy is not supported. Only one level. I was trying to do the same as you (see my comments above)

klindklind:
@egarbi thanks for the confirmation! Multi-level is mandatory for me so I need to look some alternatives at this point.

bubalush:
Are you planning to support multi-level hierarchy?

aavileli:
AWS recommends using multi-level hierarchy to store app KV. Is this going to be supported or we need to use something like

Issue #34, opened 2017-10-05
Docs request: describe relevant AWS limits


maplebed:
I set this up, and the first time we did a quick(ish) roll of one service, we got throttled by AWS. In the log we see `ThrottlingException: Rate exceeded`, which is thrown from the call to `s.svc.DescribeParameters`. 

The problem is that I can't figure out which limit is being applied in order to ask AWS to raise it, or if I've set up chamber wrong in some way that is causing it to make more calls than it should.

When trying to recreate this on my laptop, I launched 10 parallel versions of ` while :; do ./chamber exec dev -- env ; done` and triggered the throttle after ~20 requests. This seems very low.

Could you describe which AWS limits are relevant to what parts of chamber's execution, and when we should expect to hit each? If you have any idea which limits they're happy to adjust and which we need to work around that would also be valuable.

PR #36, opened 2017-10-10
add --retries/-r root flag and configure the SSM client to retry

toshok:
we at honeycomb are running up against aws throttling DescribeParameter api calls (issue #34)

This diff adds a flag for max-retries (`--retries`/`-r`).  The default value is 10, which seems to fix things for 15 simultaneous runs of `while :; do chamber exec dev -- echo hi ; echo bye; done`.

The flag is attached to the `RootCmd`, so invoking it will look something like:

```
$ chamber --retries 100 exec something -- command
$ chamber -r 100 exec something -- command
```

`numRetries` is passed to all `store.NewSSMStore` calls, which doesn't feel very golang-like.  Maybe I should add a `store.StoreConfig` struct?  Let me know!

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -51,7 +51,7 @@ func execRun(cmd *cobra.Command, args []string) error {
 	}
 
 	env := environ(os.Environ())
-	secretStore := store.NewSSMStore()
+	secretStore := store.NewSSMStore(numRetries)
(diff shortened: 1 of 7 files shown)


dfuentes:
Looks great, thanks for the improvement!

PR #39, opened 2017-10-25
use syscall.Exec instead of exec.Command

### LINKED ISSUE #37: consider calling syscall.Exec
execRun does a ton of work to pass through stdin, stdout, signals, return codes, etc from a single child process.

Instead of doing that I am wondering if it would be easier to simply call `syscall.Exec` with the arguments you want. That would replace the current process with the child process, and you wouldn't need to worry about signals or inputs/outputs being passed through correctly.

https://golang.org/pkg/syscall/#Exec

achille-roussel:
Fixes #37 

Seems like it works, I would take some advises on how to better test this:
```
$ aws-vault exec stage -- ./chamber exec service -- env
TERM_PROGRAM=iTerm.app
EDITOR_DEFAULT=subl
TERM=xterm-256color
SHELL=/bin/bash
...
```
(not sure what env var we're supposed to see getting set here, also I just learned by default editor is sublime... what a shame)

Here are the changes I made:
- move the fallback execution logic (copy/pasting) to `exec_default.go`
- add the execution based on syscall.Exec in `exec_unix.go` with build tags to use it on unix systems
- change the default execution logic to forward all signals from 1 to 31, so on platforms other than linux and darwin we do our best not to mask signals intended to the program ran by chamber.

Please let me know if anything should be changed.

### DIFF
--- a/cmd/exec_default.go
+++ b/cmd/exec_default.go
@@ -21,32 +21,22 @@ func exec(command string, args []string, env []string) error {
 	ecmd.Stderr = os.Stderr
 	ecmd.Env = env
 
-	signals := make([]os.Signal, 30)
-	for i := range signals {
-		signals[i] = syscall.Signal(i + 1)
-	}
+	signal.Notify(sigChan)
 
-	// Forward SIGINT, SIGTERM, SIGKILL to the child command
-	sigChan := make(chan os.Signal, 1)
-	signal.Notify(sigChan, signals...)
+	if err := ecmd.Start(); err != nil {
+		return errors.Wrap(err, "Failed to start command")
+	}
 
 	go func() {
 		sig := <-sigChan
-		if ecmd.Process != nil {
-			ecmd.Process.Signal(sig)
-		}
+		ecmd.Process.Signal(sig)
 	}()
 
-	var waitStatus syscall.WaitStatus
-	if err := ecmd.Run(); err != nil {
-		if err != nil {
-			return errors.Wrap(err, "Failed to run command")
-		}
-		if exitError, ok := err.(*osexec.ExitError); ok {
-			waitStatus = exitError.Sys().(syscall.WaitStatus)
-			os.Exit(waitStatus.ExitStatus())
-		}
+	if status, err := emcd.Wait(); err != nil {
+		emcd.Signal(os.Kill)
+		return errors.Wrap(err, "Failed to wait for command termination")
+	} else {
+		waitStatus := status.Sys().(syscall.WaitStatus)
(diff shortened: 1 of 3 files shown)


cmd/exec_default.go
```diff
@@ -0,0 +1,52 @@
...
+	ecmd.Env = env
+
+	signals := make([]os.Signal, 30)
+	for i := range signals {
+		signals[i] = syscall.Signal(i + 1)
+	}
+
+	// Forward SIGINT, SIGTERM, SIGKILL to the child command
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, signals...)
```
kevinburke: I believe if you call `signal.Notify` with no arguments, it listens for all signals...

cmd/exec_default.go
```diff
@@ -0,0 +1,52 @@
...
+
+	go func() {
+		sig := <-sigChan
+		if ecmd.Process != nil {
```
kevinburke: There's a data race here between `ecmd.Run` and checking `ecmd.Process` here. See https://github.com/99designs/aws-vault/pull/162 for the fix; you want to call Start() and then Wait() instead of Run().

cmd/exec_unix.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func exec(command string, args []string, env []string) error {
+	argv0, err := osexec.LookPath(command)
+	if err != nil {
+		return err
+	}
+
+	argv := make([]string, 0, 1+len(args))
+	argv = append(argv, command)
```
kevinburke: Alternatively, argv[0] = command, copy(argv[1:], args...)

modax:
Hello,

any news on this?

achille-roussel:
I missed the reviews, I’ll address them tomorrow so we can move forward.

modax:
Hi @achille-roussel,

do you need some help? This fix is kind of critical. Otherwise it is not possible to rely on application being shutdown gracefully.

cmd/exec_default.go
achille-roussel: Good call 👍

cmd/exec_unix.go
achille-roussel: A bit too much magic, I'll keep the sequential form if you don't mind (it just saves one dynamic memory allocation which is blanked out right away when `syscall.Exec` is called.

achille-roussel:
@kevinburke all your comments should have been addressed.

@dfuentes when you get a chance can you give this a look?

dfuentes:
Looks great, thanks!

Issue #20, opened 2017-08-21
Feature Request: chamber delete(rm) <service> or <service> <key>


sc-chad:
Basically do the opposite of write with the added functionality of deleting an entire service if applicable.

IE: `chamber delete service` would delete all keys with that namespace
or
`chamber delete service key` would delete just that specific key

Addition it will have to delete in batches of 10 due to the AWS constraint: ```Member must have length less than or equal to 10```

jleclanche:
Seconded. Is there really no way to delete something once it's been added?

dfuentes:
https://github.com/segmentio/chamber/pull/40 added the ability to delete individual keys via `chamber delete <key> <service>`.  Available in release v1.11.0+

nickatsegment:
@thattomperson 	That solution looks fine to me. I would certainly accept a PR for a `delete-service` command, but it's not top priority for us atm

PR #40, opened 2017-10-26
Delete entries

smulube:
This PR addresses an issue I had where an incorrect secret key was added to a service and there seemed to be no way of removing this key once it was there. 

It adds the following command:

```bash
$ chamber del service key
```

command, which just attempts to delete the specified key for that service using `ssm.DeleteParameter`.

This PR relates to #20, but is more limited in scope in that it does not attempt to address deleting an entire service, just removing a single bad key from the service. Would be interested in any feedback on this approach. I have tried out the functionality locally, and it seemed to work, but I don't fully understand the underlying storage mechanism on AWS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,16 @@ that is unneeded. Parameter store automatically versions secrets and passing
 the `--version/-v` flag to read can print older versions of the secret. Default
 version (-1) is the latest secret.
 
+
+### Deleting
+```bash
+$ chamber delete service key
+```
+
+`delete` provides the ability to remove a secret from chamber permanently,
+including the secret's additional metadata. There is no way to recover a
+secret once it has been deleted so care should be taken with this command.
+
 ### AWS Region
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -117,6 +117,16 @@ that is unneeded. Parameter store automatically versions secrets and passing
 the `--version/-v` flag to read can print older versions of the secret. Default
 version (-1) is the latest secret.
 
+
+### Deleting
+```bash
+$ chamber del service key
```
dfuentes: Do you mind renaming this to `chamber delete`?

smulube:
Hi @dfuentes - sorry for delay in getting back to this. I saw your feedback but then it escaped my mind till now. I think the above PR addresses the changes you requested, and the tests pass so hopefully I didn't mess anything up in rebasing my work on master.

dfuentes:
Looks great, thanks for following up!

PR #42, opened 2017-10-30
exec: utilize `ErrCommandMissing`

stephenmathieson:
Debugging something and noticed this. Figured I'd send a quick patch to fix =]


dfuentes:
whoops, good catch

PR #55, opened 2017-12-07
Implement persistent cmd flag to expand list output with values

xhochn:
This is a tiny improvement to list secrets with values


dfuentes:
Looks great, thanks for the contribution

PR #57, opened 2017-12-15
 Add export command

modax:
It allows to export parameters in various formats. In this comment, the
following formats are supported:

* json
* java-properties (.properties)
* csv
* tsv

Depends on https://github.com/segmentio/chamber/pull/56 (though could be refactored not to depend on it)


dfuentes:
Looks great, thanks again

PR #58, opened 2017-12-22
Base path support

modax:
This PR introduces kind of two subfeatures but they are interconnected that I submit them via single PR. Check in each commit for more information about them.

* Support base path for services.
* Add support for configuration in base path.


dfuentes:
I'm not sure we're going to go forward with this kind of feature.  Can you explain more about your use case?

modax:
Hello,

Sure. I can explain it.

Basically, we have been looking for a tool to manage and share (easy for users to see and change them) configuration easily on AWS. We are using terraform for provisioning but it does not really cut it for secrets (due to secret leaks to state file and other small inconveniences),

For development, we have a single AWS account which is used by a lot of projects. Some projects want to use their own KMS key for SSM parameters (because that's how it is going to be used in production) or they might prefer USE_PATHS approach to a standard dot-separated approach (which kind of makes sence http://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameters-by-path.html and respective SDK APIs exist). We also have a team of devops people who help with infrastructure stuff but jump from one project to another many times in a day. So while chamber behaviour can be changed by environment variables, this is cumbersome and needs to be communicated to users in projects and its very incovenient for devops who jump from project to project. Practically, the only solution is to wrap chamber around with scripts which set appropriate environment variables etc. so users can be 100% sure chamber works correctly for the project. But then again, I see a wrapper script as kind of a hack. It needs to be shipped to users and to application deployments and it is more error-prone. Then again, we use terraform and why not to be able to set up chamber with terraform once, make the config readily available for all users/deployments and forget and the problem. This chamber config in SSM parameter feels like a perfect solution. Users only need to know a base path (which remaining static and is very unlikely change separately, e.g. /test/my-awesome-project) and they are ready to use chamber. For ideal UX, chamber should gain a command to list services but might be difficult in for non-USE-PATHS and it is a topic of other PR.

Another big reason is that "service" feels like a name of microservice/component in the project base allows it to become scoped to the project, no need to worry about collisions with other projects (or use special formatting/prefixing for service names). In this sense, base is more like a namespace, I'm open to suggestion for a better naming :-) What is more, sometimes we need to spawn multiple instances of the same project in the same AWS account with independent configurations. Right now it is not that easy to do with chamber (messing with service name is possible though but feels hacky) since AWS account is the other only "dimension" beyond service name.

Last but not least, base is optional in this implementation, I don't see much harm which it could do. Though I agree that there are two overlapping concepts in this PR (configuration sharing via SSM parameter and namespace for services) but I didn't want to complicate things and reused a single value (`base`) for both.

Feel free to ask any more questions. I looking forward to this being in the mainline :-)

P.S. I will fix conflicts soon

modax:
Conflicts against current master resolved

modax:
@dfuentes: Hi! Has your view on this changed after my explanation so that you are willing to eventually merge it?

I will resolve conflicts later after other PRs are merged since here will be even more conflicts.

DanielRedOak:
I'm definitely interesting in some of this functionality and what really caught my eye was the "spawn multiple instances of the same project in the same AWS account with independent configurations" bit as we do something very similar. Unfortunately I don't think this PR would solve our issues right out of the box though as it assumes a single base_path for all.

Really what may solve more of the same subset of issues would be full multi-level hierarchical support which I've seen others asking for in other issues/PRs. Base_path is cool and would allow for some of this work, but with full multi-level you could emulate the base_path support and more.

For our multi instance deployment, I could see taking advantage of the multi-service specification on the command line. The first service would be the main environment with the set of defaults, the second could contain some overrides that are specific to the deployment/instance I am going to run.

Something like this:
`chamber exec /testing/myapp /testing/qa1/myapp`

With secrets further down in the hierarchy, lining up to the file/configuration set they belong to if it is a larger legacy style monolithic application.

```
/testing/myapp/externalapi1/<secrets here>
/testing/myapp/internalapi/<secrets here>
... etc.
```

If multi-level was implemented, I would think something like this would work to fit your use case of needing a base_path

```
export BASE_PATH=/something/here
chamber exec $BASE_PATH/my_app
```

Unfortunately it sounds like Segment doesn't use a hierarchy, so this implementation would likely need to come from someone externally :)

dfuentes:
Thanks for the feedback and insight (from both of you) on your use cases.  For now, we are not going to go forward with this change.

One of our goals with chamber is to keep the design as simple as possible.  To that end we've tried to provide just the primitives needed for handling simple secret storage.  This change expands the scope of chamber further then we're comfortable with.

For managing chamber settings from app to app, I don't see wrapper scripts as a hack.  I think it's pretty common for deployed applications to have script entrypoints, and prefer this to having chamber manage remote state.

We currently use @DanielRedOak's model for shared secrets/configuration between multiple services (which is why we added specifying multiple services on calls to exec).  Besides the AWS API's for working specifically with path's, I'm not sure I see the difference between:

`chamber exec /testing/myapp /testing/qa1/myapp`
and
`chamber exec testing-myapp testing-qa1-myapp`

which is supported and works right now.  If the reason is wanting to use chamber in some places, and AWS path-based API calls elsewhere, I could see why this is less desirable.  This sort of mixed chamber/raw AWS API mode is not something we really designed around.

modax:
Hi, @dfuentes,

this is unfortunate news to be honest :/ The way we saw chamber was as an ultimate tool for configuration management (e.g. in combination with terraform `aws_ssm_parameter` resource for non-secrets), not just secret management. awscli SSM interface is just terrible and `chamber` seemed like a breath of fresh air. Not sure if we are going to maintain a fork or not yet but I still would like push as many changes as possible since it would benefit everybody.

1) For example, would you accept the patch to add support for `/` in services names if you dislike base concept? `/` is a natural separator for such things. As far as I see there is no technical limitations for that and code won't look terrible due to news [BeginsWith parameter filter which AWS API currently supports](https://github.com/segmentio/chamber/pull/58/files#diff-0872b2d0c54e370d387fa05379be3564R313) (as seen in this PR)

2) Maybe if not full config management with save as all those bells and whistles, would you accept a `config show` (to confirm what settings are in use) and maybe config file parameter (in addition to envvars)? It might be useful to see what configuration settings chamber will use.

PR #61, opened 2017-12-27
Do not query EC2 metadata API if region is in ~/.aws/config.

modax:
Previous code would still query EC2 metadata API if region is found in
~/.aws/config. This might make chamber run slow (10 secs delay on my
machine) on non-EC2 instances due to extra HTTP request to
169.254.169.254 which does not exist.

What is more, let AWS SDK read region from AWS_REGION environment
variable, do not attempt to do this manually.


dfuentes:
Looks good, I've written some unit tests for this locally, I will add those after I merge this

PR #62, opened 2017-12-27
Add suport for --singleline option in write command

pangorgo:
After using `chamber` for some time I find that most of time I put single line secrets there. My other observation is that it becomes error prone (especially for new users) to remember about pressing `Ctrl+D` at the end of string.

That's why I propose to add `--singleline` option to save parameter just after hitting enter. 

Another use case for that is copying parameters between services:
```chamber read -q service1 parameter | chamber write -s service2 parameter -```

Without this feature, provided example adds trailing `\n` at the end of copied value.

### DIFF
--- a/cmd/write.go
+++ b/cmd/write.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bufio"
 	"io/ioutil"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


dfuentes:
Looks good, thanks for the feature

PR #63, opened 2017-12-28
Allow dots in secret keys.

modax:
Dot is pretty common in java configuration (property) files. There is no
technical reason why dot should not be allowed. On the other hand,
service names should not be allowed to have dots.

What is more, refactor current key parsing a bit and move it to ssmstore
where it really belong (IMHO).


dfuentes:
Would you mind writing some tests for this?  Mainly around the validation for path/non-path usage.

modax:
Sure, I will try

Issue #21, opened 2017-08-22
cli displays lowercase even though its stored in uppercase


stevenmiller888:
![image](https://user-images.githubusercontent.com/7015626/29587098-aff99ffa-8741-11e7-815b-1a26663c81b6.png)

It's confusing that the env var name is stored in uppercase but displayed in lowercase

dfuentes:
It is stored in lower case as well.  This is to prevent people from accidentally writing two similar secrets (AUTH0_SECRET and auth0_secret), which would be ambiguous when we convert to env vars for `exec`.

stevenmiller888:
There should at least be a message in the output indicating this. Otherwise I guarantee someone else will run into this and get confused

PR #66, opened 2018-01-03
add support for case-sensitivity in secret keys

sjauld:
Currently, the keys for secrets are down-cased before storage, and then upcased during the `chamber exec` command.

However, ENVs are case sensitive. When attempting to use chamber in conjunction with Terraform.
 this causes problems since [terraform consumes ENVs in the format TF_VAR_my_variable](https://www.terraform.io/docs/configuration/variables.html).

#21 mentions that the current behaviour is deliberate, however it would be nice (critical?) to have the option to run in case-sensitive mode. IMHO it would be even better for case-sensitive mode to be the default - if you want to export ENVs in uppercase, why not store them in uppercase too?

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -26,7 +26,14 @@ func delete(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "Failed to validate service")
 	}
 
-	key := strings.ToLower(args[1])
+	var key string
+
+	if caseSensitive {
+		key = args[1]
+	} else {
+		key = strings.ToLower(args[1])
+	}
+
 	if err := validateKey(key); err != nil {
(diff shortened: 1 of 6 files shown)


dfuentes:
I stand by the response in #21 still.  Environment variables are case insensitive in some environments, and not in others (like windows).  I rather be consistent across the board.

It's unfortunate that terraform works this way, but there are some workarounds you could use if you wrap your invocation in a script:
- Re-export env vars in the right form
- Use `-var` flags in your calls to terraform

sjauld:
Wait, so you won't even consider opt-in support for case sensitivity?

bee-keeper:
Would you consider a PR for a flag which would allow for case sensitivity?

PR #67, opened 2018-01-03
add log on error

anoonan:
We are seeing issues where integration containers are exiting with exit code 1 and no logs. We have logs in place on start up that are not showing up and this is the only thing that happens before that. Adding logs here to hopefully get insight into why the containers are exiting.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/segmentio/events"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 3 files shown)


f2prateek:
Wouldn't the output of `cmd.Usage()` already be captured in CloudWatch?

anoonan:
@f2prateek only if it's a usage error, but good point. I'll wrap in an else statement

dfuentes:
The error will be printed without this change due to the implementation of `Execute()` (https://github.com/spf13/cobra/blob/master/command.go#L825).

If you are seeing an exit code of 1 with no error, that is likely your child process just exiting with status code 1 and not logging anything itself.  WIth the new `exec` implementation, we won't even reach that log statement if your code is executed (because https://github.com/segmentio/chamber/blob/master/cmd/exec_unix.go#L21 hands off execution to the child)

anoonan:
@dfuentes ahh, got it. thanks!

PR #70, opened 2018-01-19
Make the service path support variable depths

imjoshholloway:
This PR makes it possible to have the service have a variable depth for the service (`<path>/<service>`). The default "." separator is also supported

e.g

```
chamber write some/path/to/service mykey myvalue
chamber write some/service myotherkey myothervalue

chamber list some/path/to/service
Key             Version         LastModified    User
mykey    1               01-19 13:34:56  arn:aws:iam::0000000000:user/imjoshholloway

chamber list some/service
Key             Version         LastModified    User
myotherkey    1               01-19 13:34:56  arn:aws:iam::0000000000:user/imjoshholloway

chamber list some
Key             Version         LastModified    User
```

Happy to write some tests if needed but wanted to guage the desire to support this before spending too much time on it.

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -66,13 +66,12 @@ func list(cmd *cobra.Command, args []string) error {
 
 func key(s string) string {
 	_, usePaths := os.LookupEnv("CHAMBER_USE_PATHS")
+	sep := "."
 	if usePaths {
-		tokens := strings.Split(s, "/")
-		secretKey := tokens[2]
-		return secretKey
+		sep = "/"
 	}
 
-	tokens := strings.Split(s, ".")
-	secretKey := tokens[1]
+	tokens := strings.Split(s, sep)
+	secretKey := tokens[len(tokens)-1]
 	return secretKey
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,7 +12,7 @@ import (
 // Regex's used to validate service and key names
 var (
 	validKeyFormat     = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
-	validServiceFormat = regexp.MustCompile(`^[A-Za-z0-9-_\/]+$`)
+	validServiceFormat = regexp.MustCompile(`^[A-Za-z0-9-_\/.]+$`)
 
 	numRetries int
 )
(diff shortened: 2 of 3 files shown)


Joholland:
I tested this in a private build and I'm not sure if it is actually working correctly. I'll try pushing a PR out later with fixes to the regular expressions. At the moment, I get a validation error.

imjoshholloway:
@Joholland What's the validation error? We're using the following build and so far we've not seen any issues https://github.com/splunkcloud/chamber/releases/tag/temp

vood:
This looks promising. I'll try to fix the conflicts

jwilner:
Just to put in my two cents -- we'd love to use `chamber` but can't in its current form. We don't use multiple AWS accounts for namespacing, so we need to establish those namespaces within the store itself. This PR would let us do so -- so if there's anything I can do to help it along, please let me know!

vood:
I've merged this code into the latest master and built the binary. You can find it here https://github.com/people-ai/chamber/releases

vood:
Of course. My build is just a temporary workaround

dfuentes:
In order to get this merged, I think we need some more validation.  For example, I tried this build out and was able to get things in a pretty confusing state:
```bash
$ export CHAMBER_NO_PATHS=1
$ chamber write test/with/mixed key val
Error: ValidationException: Parameter name must be a fully qualified name.
	status code: 400, request id: e21e2ecf-7340-47d0-b885-b00f5ec1cc28
exit status 1
```

```
$ export CHAMBER_NO_PATHS=1
$ chamber write /test/with/mixed key val
$ chamber list /test/with/mixed
Key	Version		LastModified	User
```

Because the path based api doesn't play well with the older API, I suggest updating the validation such that service names may not contain forward slashes when `CHAMBER_NO_PATHS` is set.  Similarly, we should not allow periods in service names when `CHAMBER_NO_PATHS` is not set.

dfuentes:
I'd say "chamber is currently not usable" is a bit of hyperbole.  I'm not opposed to having this feature, but this PR in it's current state is not something I'm comfortable merging.  If the validation is fixed up and the edge cases between paths/non-paths uses are fixed, then we can get it in.

robinbowes:
I've just discovered that chamber doesn't support this which makes it unusable for our use-case too. Looks like I shall have to use raw `aws ssm` commands for now.

stefano-b2c2:
> If it helps folks, I wrote something that doesn't provide all the perks of this project but is less opinionated about namespaces and keys: https://github.com/energyhub/secretly
> 
> (all it does is aim to have a nice API for dumping values from your param store into your container)

Thank you! This is exactly what I needed.

mechanical-fish:
It appears that https://github.com/segmentio/chamber/pull/118 has built upon this PR and implemented this feature, so this specific PR can be closed. I've successfully tested variable-length service paths with the latest version of Chamber. Thanks, everyone!

PR #80, opened 2018-03-16
Use path based parameter store API by default

dfuentes:
This is a breaking change and will come with a major version bump (v2.0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ $ chamber export [--format <format>] [--output-file <file>]  <service...>
 {"key","secret"}
 ```
 
-`export` providers ability to export secrets in various file formats. The following
+`export` provides ability to export secrets in various file formats. The following
 file formats are supported:
 
 * json (default)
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -5,6 +5,12 @@ secrets in SSM Parameter Store, an AWS service for storing secrets.
 
 For detailed info about using chamber, read [The Right Way To Manage Secrets](https://aws.amazon.com/blogs/mt/the-right-way-to-store-secrets-using-parameter-store/)
 
+## 2.0 Breaking Changes
+
+Starting with version 2.0, chamber uses parameter store's path based API by default.  Chamber pre-2.0 supported this API using the `CHAMBER_USE_PATHS` environment variable.  The paths based API has performance benefits and is the recommended best practice by AWS.
+
+As a side effect of this change, if you didn't use path based secrets before 2.0, you will need to set `CHAMBER_NO_PATHS` to enable the old behavior.  This option is deprecated, and We recommend only using this setting for supporting existing applications.
```
avaskys: Might be useful to mention how to migrate existing secrets to the new path-based system. I assume it's as simple as re-saving the secret using chamber 2.0, but it would be nice to explicitly explain the process here.

dfuentes: Good call.  I've implemented a `chamber import` command now which should ease migration.  I've updated the readme for that command as well as adding directions for migrating.  Specifically, you should be able to switch to the new format using:
```bash
$ CHAMBER_NO_PATHS chamber export foo | chamber import foo -
```

avaskys: Nice! 🔥

jeffnappi:
Looking forward to seeing this released. Is there any tentative ETA?

PR #93, opened 2018-05-30
Enhancement: Adds a masking feature

natefaerber:
Here is a feature request in the form of a PR.  One thing I really love about `chamber` is that it will export type=String SSM Parameters along with type=SecureString.  That means I can put application configuration that isn't sensitive into SSM Parameter store next to the secrets I create with chamber.  Then I can use chamber to export them all.

The feature request is based on the fact that chamber will fail completely if any secret in that particular service is encrypted with a KMS key that the user doesn't have permission to use.  Not a problem if you are only using chamber for secrets.  But what if I have a mix of secrets and plain strings.  It would be awesome if QA/Dev or other non-privileged users could use chamber to export the parameters stored using chamber.  With this change, you can add the `-m` flag to effectively not fail when a secret is hit that the user can't decrypt.  It allows the encrypted string to be returned in code and simply masks the output to the user.

I think this will be great for our less privileged users to be able to verify settings in environments using the same tool Admins use to export settings.  I provide an option to customize the masking string.

Thoughts?

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -20,6 +20,7 @@ import (
 var (
 	exportFormat string
 	exportOutput string
+	exportMask = &store.MaskOptions{}
 
 	exportCmd = &cobra.Command{
 		Use:   "export <service...>",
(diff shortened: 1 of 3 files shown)


dfuentes:
Unfortunately, I don't think we are going to move forward with this feature.

While chamber does work somewhat with parameters added out of band, we don't want to take on the maintenance burden of guaranteeing that this interop will work.  By not worrying about working with parameters created or modified outside of chamber itself, we give ourselves more freedom to do things like change internal representation or support other non-parameter store backends.

PR #95, opened 2018-06-15
added list sorting options

brycehendrix:
This adds sorting options to the list command which is extremely useful in services with many secrets.

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"sort"
 	"strings"
 	"text/tabwriter"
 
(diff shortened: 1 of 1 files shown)


cmd/list.go
```diff
@@ -76,3 +94,23 @@ func key(s string) string {
 	secretKey := tokens[1]
 	return secretKey
 }
+
+type ByName []store.Secret
```
dfuentes: This seems unused, we should remove it if that's the case.

brycehendrix: It's required by the 3 lines which follow it.

dfuentes: ah, I missed line 58, didn't realize that you always sort by name first.

brycehendrix:
Not sure what the PR process here is- the PR was approved 2 weeks ago, but I don't have access to merge. Will it be merged as the next release is prepared?

nickatsegment:
Sorry for the delay! Thanks for your contribution.

nickatsegment:
Looks like the merge broke the build (probably bitrot); fixing.

nickatsegment:
Huzzah! https://circleci.com/workflow-run/aed9eb98-35cf-4704-ba3e-4646d1b86277

PR #97, opened 2018-06-23
Enhancement: Add execlocal command to use a local env file.

rickbassham:
This allows you to use a local env file for development purposes. It works the same way as `exec`, but instead of getting the parameters from AWS Parameter Store, it just reads files in `dotenv` format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ environment.
 
 ### Exec Local
 ```bash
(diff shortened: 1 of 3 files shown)


nickatsegment:
Yeah, I'm not sure I see the use case. We're trying to keep chamber tight and focussed.

Issue #92, opened 2018-05-18
feature request: `retries` to be picked up when creating ssmSession / getting NoCredentialProviders


mkj28:
aws-sdk-go has relatively short timeout when fetching credentials, which shows up as [`NoCredentialProviders` when using kube2iam](https://stackoverflow.com/questions/50415647/aws-sdk-go-getting-nocredentialproviders-when-using-kube2iam). 

The credentials are eventually passed through and you can work around by adding a `sleep 5` before chamber. But would be nice for chamber to retry (maybe with delay)?

Looking through `ssmstore` - retries are set for SSM request, but not for the SSM session itself. Would be awesome to have it there, too.

PR #111, opened 2018-08-02
S3 backend

dfuentes:
This PR adds an experimental S3 backend.  We have run into some AWS rate limiting issues around using Parameter Store at scale, even when using the newer GetParameterByPaths API.  For some very high concurrency cases, we want to explore using S3 with server side encryption, and it would be nice if we could continue to use the same chamber interface.

For storage, the S3 backend assumes a single global secrets bucket by AWS account.  The object layout looks like:

```
chamber-secrets
└── service-foo
    ├── bar.json
    ├── baz.json
    └── __latest.json
```

Where `service-foo` is a service with two secrets, `baz` and `bar`.

`__latest.json` is an index of sorts, which keeps the latest values of each secret for the service, but none of the metadata or previous versions.  This file is used by the `ListRaw` implementation to allow `chamber exec` to only use a single S3 GetObject API call.

Each individual secret file (`bar.json` for example) holds metadata about that secret including previous secret values, and is used to power the `List`, `History`, and `Read` APIs.

For IAM permissions, the idea is to configure task roles such that they have read access to only keys in the s3 bucket with prefix matching their service name.

Also included in `store/backendbenchmarks_test.go` are some simple benchmarks which compare concurrent usage performance for the SSMStore and S3Store backends.  These benchmarks are disabled during automated testing because they require infra and authed AWS users/roles.  They can be enabled by passing the `-benchmark` flag to `go test`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ If you'd like to use a different region for chamber without changing `AWS_REGION
 
 By default, chamber store secrets in AWS Parameter Store.  We now also provide an experimental S3 backend for storing secrets in S3 instead.
 
-To configure chamber to use the S3 backend, set `CHAMBER_SECRET_BACKEND` to `S3`, and `CHAMBER_S3_BUCKET` to an existing S3 bucket (by default, chamber will attempt to use a bucket named `chamber-secrets`).
+To configure chamber to use the S3 backend, set `CHAMBER_SECRET_BACKEND` to `S3`, and `CHAMBER_S3_BUCKET` to an existing S3 bucket.  Preferably, this bucket should reject uploads that do not set the server side encryption header ([see this doc for details how](https://aws.amazon.com/blogs/security/how-to-prevent-uploads-of-unencrypted-objects-to-amazon-s3/))
 
 This feature is experimental, and not currently meant for production work.
 
--- a/README.md
+++ b/README.md
@@ -196,6 +196,14 @@ for more details.
 
 If you'd like to use a different region for chamber without changing `AWS_REGION`, you can use `CHAMBER_AWS_REGION` to override just for chamber.
 
+## S3 Backend (experimental)
+
+By default, chamber store secrets in AWS Parameter Store.  We now also provide an experimental S3 backend for storing secrets in S3 instead.
+
+To configure chamber to use the S3 backend, set `CHAMBER_SECRET_BACKEND` to `S3`, and `CHAMBER_S3_BUCKET` to an existing S3 bucket (by default, chamber will attempt to use a bucket named `chamber-secrets`).
+
+This feature is experimental, and not currently meant for production work.
+
 ## Releasing
 
 To cut a new release, just push a tag named `v<semver>` where `<semver>` is a
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strings"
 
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -196,6 +196,14 @@ for more details.
 
 If you'd like to use a different region for chamber without changing `AWS_REGION`, you can use `CHAMBER_AWS_REGION` to override just for chamber.
 
+## S3 Backend (experimental)
+
+By default, chamber store secrets in AWS Parameter Store.  We now also provide an experimental S3 backend for storing secrets in S3 instead.
+
+To configure chamber to use the S3 backend, set `CHAMBER_SECRET_BACKEND` to `S3`, and `CHAMBER_S3_BUCKET` to an existing S3 bucket (by default, chamber will attempt to use a bucket named `chamber-secrets`).
```
systemizer: since bucket names are unique across all S3, we might want to not have a default. instead if it isn't set, fail with an error

cmd/root.go
```diff
@@ -62,3 +72,35 @@ func validateKey(key string) error {
...
+
+	switch backend {
+	case SSMBackend:
+		return store.NewSSMStore(numRetries)
+	case S3Backend:
+		return store.NewS3Store(numRetries)
+	}
+
+	// This line is unreachable, but necessary to satisfy the compiler
+	panic("unreachable")
```
nickatsegment: I think this can be simplified? Eg
```
    var store Store
    switch backendEnv {
        case "s3":
            store = S3Store{}
        case "ssm":
            fallthrough
        default:
            store = SSMStore{}
    }
    return store
```

store/shared.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/ec2metadata"
+	"github.com/aws/aws-sdk-go/aws/session"
+)
+
+const (
+	RegionEnvVar = "CHAMBER_AWS_REGION"
+)
+
+func getSession(numRetries int) (*session.Session, *string) {
```
nickatsegment: Looks like `numRetries` is unused, and this is a new function so we could probably lose it.

dfuentes: ah, I missed a commit (67ef9eb).  I added the retries to session creation to address https://github.com/segmentio/chamber/issues/92

cmd/root.go
dfuentes: Good call, I like that

dfuentes: ff5fb8f

README.md
dfuentes: removed in 3d34550 (code) and b4edc16 (readme)

store/s3store.go
```diff
@@ -0,0 +1,438 @@
...
+
+}
+
+func (s *S3Store) Write(id SecretId, value string) error {
+	index, err := s.readLatest(id.Service)
+	if err != nil {
+		return err
+	}
+
+	objPath := getObjectPath(id)
```
systemizer: Do we allow/care if people use "/" in their key name? Not sure if there's validation elsewhere, but I think if someone has "/" in the key name, it will make a new folder?

dfuentes: We do validation in the CLI code (https://github.com/segmentio/chamber/blob/master/cmd/write.go#L38 for write, but similar for read/etc)

PR #118, opened 2018-08-29
Support Variable Depths and improve regex

pecigonzalo:
Based on the work of #70 but improving the regex and adding a separate validator for `/` paths.

Given we have no tests, I have left some comments explaining how the validation is expected to work, but I can easily remove it if you believe its too much noise/ugly. (which it is, but I dont know how else to document this in this repo)

Relates to: #70 

```go build -o dist/chamber-local-linux-amd64
# Test Key Validation
## Works as expected
./dist/chamber-local-linux-amd64 write validKey foo secret
./dist/chamber-local-linux-amd64 write validKey foo.bar secret
./dist/chamber-local-linux-amd64 write validKey foo. secret
./dist/chamber-local-linux-amd64 write validKey .foo secret
## Fails as expected
./dist/chamber-local-linux-amd64 write invalidKey /foo secret
./dist/chamber-local-linux-amd64 write invalidKey foo//bar secret
./dist/chamber-local-linux-amd64 write invalidKey foo/bar secret
## Displays all keys
./dist/chamber-local-linux-amd64 list validkey
## Shows all envs correctly
./dist/chamber-local-linux-amd64 exec validkey -- env | grep FOO


# Test Service Validation CHAMBER_NO_PATHS=1
## Works as expected
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 write validKey foo secret
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 write valid.Key foo secret
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 write valid.Key.Service foo secret
## Fails as expected
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 write validKey. foo secret
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 write .validKey foo secret
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 write valid/Key foo secret
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 write valid..Key foo secret
## Displays all keys
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 list validKey
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 list valid.Key
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 list valid.Key.Service # Recurses to `valid.Key.Service`
## Shows all envs correctly
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 exec validkey -- env | grep FOO
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 exec valid.Key -- env | grep FOO
CHAMBER_NO_PATHS=1 ./dist/chamber-local-linux-amd64 exec valid.Key.Service -- env | grep FOO

# Test Service Validation (using / and PATHS)
## Works as expected
./dist/chamber-local-linux-amd64 write validKey foo secret
./dist/chamber-local-linux-amd64 write valid/Key foo secret
./dist/chamber-local-linux-amd64 write valid/Key/Service foo secret
## Fails as expected
./dist/chamber-local-linux-amd64 write validKey/ foo secret
./dist/chamber-local-linux-amd64 write /validKey foo secret
./dist/chamber-local-linux-amd64 write valid//Key foo secret
./dist/chamber-local-linux-amd64 write valid.Key foo secret
## Displays all keys
./dist/chamber-local-linux-amd64 list validKey
./dist/chamber-local-linux-amd64 list valid/Key # Does not recurse to `valid/Key/Service`
./dist/chamber-local-linux-amd64 list valid/Key/Service
## Shows all envs correctly
./dist/chamber-local-linux-amd64 exec validkey -- env | grep FOO
./dist/chamber-local-linux-amd64 exec valid/Key -- env | grep FOO
./dist/chamber-local-linux-amd64 exec valid/Key/Service -- env | grep FOO```

### CODE REVIEW

**Comment by pecigonzalo** [comment]
Oh I did not notice the test there, my bad ill add them

**Comment by pecigonzalo** [comment]
@nickatsegment I noticed that `store/ssmstore_test.go` is not actually doing validation of the regexs, and that fields are also not validated on `write` for the `SSMStore` only on `list` which makes the testing a bit more complex.
I think its out of the scope of this PR, but probably it should also validate on `write`

**Comment by nickatsegment** [comment]
> Oh I did not notice the test there, my bad ill add them

No worries. :)

> I noticed that store/ssmstore_test.go is not actually doing validation of the regexs, and that fields are also not validated on write for the SSMStore only on list which makes the testing a bit more complex.

I was thinking just simple unit tests for `validateService` and `validateKey` against your valid/invalid lists. Though testing that write/list/etc. validate in the same way could be valuable, but sort of duplication.

> I think its out of the scope of this PR, but probably it should also validate on write

Probably :) Wanna cut a bug for that?

**Comment by pecigonzalo** [comment]
@nickatsegment Added the tests, which found some small issues (which I believe would not happen anyway because of how the rest of the components work) and fixed it.

Added tests for `ssm_store` and also to `root`, let me know if that is ok.

**Comment by nickatsegment** [comment]
> Added tests for ssm_store and also to root, let me know if that is ok.

Cool. Out of scope, but at some point it'd be nice to centralize this validation.

Tests pass on my machine, so LGTM. I'm going to loop in some of the other maintainers before merging: I'm new :)

**Comment by pecigonzalo** [comment]
I agree. I mentioned on the linked #119 that we should do that as well, not only for the test but I believe it is probably for the best that both validations are the same.

File: cmd/root.go [hunk 1]
```diff
@@ -11,8 +11,9 @@ import (
 
 // Regex's used to validate service and key names
 var (
-	validKeyFormat     = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
-	validServiceFormat = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
+	validKeyFormat         = regexp.MustCompile(`^[\w\.]+$`)
```
AndreyMarchuk: does not support dashes and underscores

cmd/root.go
```diff
@@ -11,8 +11,9 @@ import (
...
+	validServiceFormat     = regexp.MustCompile(`^[\w]+(\.[\w]+)*$`)
+	validServicePathFormat = regexp.MustCompile(`^[\w]+(\/[\w\.]+)*$`)
```
AndreyMarchuk: does not support dashes and underscores

cmd/root.go
```diff
@@ -11,8 +11,9 @@ import (
 
 // Regex's used to validate service and key names
 var (
-	validKeyFormat     = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
-	validServiceFormat = regexp.MustCompile(`^[A-Za-z0-9-_]+$`)
+	validKeyFormat         = regexp.MustCompile(`^[\w\.]+$`)
+	validServiceFormat     = regexp.MustCompile(`^[\w]+(\.[\w]+)*$`)
```
AndreyMarchuk: does not support dashes and underscores

cmd/root.go
AndreyMarchuk:
```
^[\w\.-]+$
^[\w-]+(\.[\w-]+)*$
^[\w-]+(\/[\w\.-]+)*$
```

maartenvanderhoef:
Hi everyone, I'm late in the game here, big thumbs up for everyone contributing! I see quite some effort was made already into giving the service a different path which is just what I was looking for 👍 

My idea (https://github.com/segmentio/chamber/issues/128) however is different from implementation and would like know what you think.. I suggested having a `SSM_PATH_NAMESPACE` env var to be prepended in front of the service.

This would keep the Dockerfile cleaner and I think a bit more context agnostic. If the same docker is deployed in the same region using a different path (think of migrations) with the current PR, it would look like:

`CMD chamber exec ${PREPEND}/Service`

Having chamber interpolating SSM_PATH_NAMESPACE as ENV var, it would look like this:

`CMD chamber exec service`

The other advantage of having a single depth service with a prepend is that this leaves the possibility for a later moment to create variables with variable depth .. One could do a replace on / with _ and you can then have

<pre>
/service/monitoring/logging_this_app_key     MONITORING_LOGGING_THIS_APP_KEY
/service/monitoring/data_api_key             MONITORING_DATA_API_KEY
</pre>

Thanks!

pecigonzalo:
@maartenvanderhoef I think your requirement it covered by this PR, we could also add the ENV VAR, but Isn't it easier to just interpolate your prefix on your CMD line?

cmd/root.go
pecigonzalo: @AndreyMarchuk Good catches, I updated the test cases and the regex both for SSM and for the CMD. I will put it as a fixup, so you can see the changes, once we are ready to merge I can squash them.

pecigonzalo: One slight comment, I left some of the old functionality where a `/./this` path/key is allowed, which I believe should be fine and it leaves on the user's hands using a proper naming convention.

pecigonzalo:
Not that `fixup!` commits are there just for visibility of the change right now, ill squash them once this is ready to merge.

maartenvanderhoef:
Hi Gonzalo, that's right, it's covered 👍. I think it would be great if an ENV VAR could work as well.. I think that it is more clean looking compared to Dockerfiles which have to 'interpolate' a variable in the entrypoint or cmd.

pecigonzalo:
I would keep that as a separate PR as it is a different feature, and it can be done sort of easily once this is merged.

cmd/root.go
AndreyMarchuk: Awesome, thanks! I like that we are doing a quick progress.

pecigonzalo: This should be ready to review now.

pecigonzalo:
@systemizer @AndreyMarchuk @nickatsegment given the approves so far, I squashed the fixups, so this is ready to merge.

nickatsegment:
@pecigonzalo Thanks! Merged. FYI we almost always do a squash and merge

PR #122, opened 2018-09-03
Stop paginating after finding a matching parameter.

jmcarp:


jmcarp:
Ping @nickatsegment 👋

systemizer:
Nice catch! Looks like a bug to me.  

However, we don't want to return true if we don't find the parameter. I think we want to keep ~lastpage so that we keep iterating in case we don't find the parameter on the first page

jmcarp:
Looking at the sdk [docs](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/making-requests.html#using-pagination-methods), pagination continues until either the callback returns `false` or there aren't any more pages to fetch. So I think this patch is correct--we bail out (return `false`) if we find the matching parameter, else continue paging if there are more pages (return `true`). We could continue returning `!lastPage`, but that wouldn't actually change the behavior of the function. IMO this way is clearer (the `lastPage` parameter isn't really useful to us), but happy to revert that part if you prefer.

Does that make sense? Sorry if I'm misunderstanding.

systemizer:
Oh I misread. That works, too. 

The way I saw it is: you are fixing the case to break out of the pagination when we find the parameter. Otherwise the behavior of returning `!lastpage` and `true` is essentially the same thing (keep iterating over all pages until a break case).

This looks good to me. thanks for the contribution!

PR #124, opened 2018-09-04
Return original error on reading parameter.

### LINKED ISSUE #104: Chamber erroneously returns "secret not found" when access is denied
If the SSM API returns an access denied error due to insufficient permissions to access a secret, Chamber will erroneously return a `secret not found` error instead of the permission denied error from the API.  This misleads the operator as to what the actual problem is.

jmcarp:
If there's no matching parameter, the api returns an empty list, in which case we return `ErrSecretNotFound`. Otherwise we should return the original error.

[Fixes #104]

### DIFF
--- a/store/ssmstore.go
+++ b/store/ssmstore.go
@@ -181,7 +181,7 @@ func (s *SSMStore) readLatest(id SecretId) (Secret, error) {
 
 	resp, err := s.svc.GetParameters(getParametersInput)
 	if err != nil {
-		return Secret{}, ErrSecretNotFound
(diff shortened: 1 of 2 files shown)


nickatsegment:
@jmcarp Seems sensible to me. Were you hoping to get a specific error from AWS?

jmcarp:
@nickatsegment: yep, I ran into the same issue as #104: when I can't read a secret due to permissions, I want to see an `AccessDeniedException` instead of "secret not found".

PR #125, opened 2018-09-04
Circle 2.0

nickatsegment:


.circle/config.yml
```diff
@@ -0,0 +1,43 @@
...
+      - checkout
+      - attach_workspace: { at: . }
+      - run:
+          name: Release
+          command: |
+            make release
+
+workflows:
+  version: 2
+  test-deploy:
```
systemizer: nit: name is confusing

nickatsegment: Agreed. Fixed.

Issue #115, opened 2018-08-24
Feature Request: Ability to run chamber exec without credentials with environment variable


albertpaulp:
I'm using chamber for production, staging environments now. I use chamber along with Docker. I use same docker image for Production, Staging and development/local, however there is a problem with this setup. chamber errors out if I run `chamber exec` without setting it up initially. I don't want to store my development/local environment secrets as it change often and other developers use their own secrets.

Proposed solution is using a environment variable like `CHAMBER_DEVELOPMENT_MODE` to change Chamber behaviour, so that it won't fail and execute the command passed in even if chamber not configured.

I'm a ruby developer and I'm completely new to Go else I would've submitted a MR along with this. It'll be very helpful if we can have this feature.

nickatsegment:
@albertpaulp what set up are you talking about specifically? The KMS key? AWS auth stuff?

albertpaulp:
@osterman Sorry for radio silence, yes it does.

PR #127, opened 2018-09-05
+ None Backend

maartenvanderhoef:
# Adding None Backend
Adding a None Backend to make it easy to disable any kind of backend lookups. In some cases, especially during migrations it is not possible to simply disable Chamber by changing the command. To then put chamber inside a wrapper script is also quite cumbersome.

My suggestion is to add a None backend, which would simply passthrough ENV when the ENV variable `CHAMBER_SECRET_BACKEND` has been set to `NONE`.


osterman:
@maartenvanderhoef is this also related to this? https://github.com/segmentio/chamber/issues/115

maartenvanderhoef:
@osterman 100%

nickatsegment:
Good stuff, how do we feel about the name? It's fine as is, but how about NoopBackend? Null? Fake? Mock?

systemizer:
I think this is fine. I'm curious about the use cases for needing this functionality. 

For example, when you are running migrations, what is forcing you to run chamber when doing this? Maybe there is additional tooling that just requires you to run chamber nonetheless? I'm curious what the forcing function is here.

maartenvanderhoef:
@systemizer I think you have migrations in mind like database migrations. what I meant is; migrating docker services from A to B. I explained it here: https://github.com/segmentio/chamber/issues/143

@nickatsegment Null was my second thought. Let me know what naming you like and I'll change it.

systemizer:
@maartenvanderhoef thanks for the explanation. I can now see how using the same docker images w/ chamber across cloud providers can lead to difficulties.

I'm also leaning towards `NullStore`. how about we change None --> Null and then we merge?

maartenvanderhoef:
Apologies, I made https://github.com/segmentio/chamber/pull/149 as I couldn't figure out how to leave my merge issues.

Issue #117, opened 2018-08-28
Using chamber with terraform tip


conzy:
Just want to say thank you for chamber. It has made adopting aws parameter store a breeze for us.

Terraform expects variables to be passed via the environment like this `TF_VAR_the_variable`

I have lots of terraform modules that expect a database password or some secret. I put the secrets in parameter store under a `/terraform/` namespace. I can then inject secrets into terraform at run time with this simple script called `terrachamber`

```ruby
#!/usr/bin/env ruby

# Takes a terraform / terragrunt command as an argument.
# Uses a system call to grab terraform secrets from AWS parameter store via chamber
# Injects those secrets into the environment in the terraform variable format i.e TF_VAR_the_variable
# Performs a system call of the argument array that is passed
#
# Usage Examples
#
#  terrachamber terraform apply
#  terrachamber terragrunt apply --terragrunt-source-update
#  aws-vault exec dev-admin -- terrachamber terraform apply
#

require 'json'

JSON.parse(`chamber export terraform`).each {|k,v| ENV["TF_VAR_#{k}"] = v}
system(*ARGV)
```

This may help someone else. But this is also a feature request :)

If you could chose the output format of chamber that would be wonderful. e.g I could do the above natively with chamber if it supported a key prefix and if it supported the ability to not capitalise the environment variables.

Something like:

`chamber exec --prefix TF_VAR_ --no-capitalise namespace -- terraform apply`

That would take my secrets:

```
/terraform/database_password
/terraform/big_secret
```
And export them like this:

```
TF_VAR_database_password
TF_VAR_big_secret
```

This may be beyond the scope of chamber. But it already supports json output etc.

Thanks!

nickatsegment:
Sounds reasonable to me. I wonder if a general `--format` flag with templating would work, so you could do something like
```
chamber exec --format 'TF_VAR_{{.}}'
```
where the default is `{{. | upper }}`

systemizer:
Hey @conzy !

Thanks for your thoughts here. As mentioned in a [similar PR](https://github.com/segmentio/chamber/pull/130#issuecomment-419538123) , we have decided variable transforms are outside the scope of chamber. We recommend applying formatting either within your code or via a intermediate formatting tool.

Let me know if this makes sense. Thanks again!

nickatsegment:
Sorry for leading everybody on :)

calebgosnell:
[...]

PR #130, opened 2018-09-07
Issue 117 exec format

jmcarp:
I added a collection of useful template functions from github.com/Masterminds/sprig because `text/template` isn't very useful on its own.

cc @nickatsegment @conzy

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -1,13 +1,18 @@
 package cmd
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"strings"
+	"text/template"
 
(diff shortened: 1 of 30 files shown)


systemizer:
hey @jmcarp !

We talked about this within our team and decided that this functionality is outside the scope of chamber. Our vision for chamber is to keep the interface small and focused solely on managing secrets. 

This isn't to say transforms aren't useful.  We believe transforms should be done outside of chamber, whether that's done in the application code, or some intermediate formatting tool.

jmcarp:
Sounds reasonable @systemizer. You might want to close #117 in that case--it's still open and tagged with "help wanted".

PR #131, opened 2018-09-10
Print env var keys on exec with a global`-v` flag

YarekTyshchenko:
It's useful to know the exact keys that are picked up by Chamber as that
information is not otherwise available anywhere else

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -39,6 +39,7 @@ func execRun(cmd *cobra.Command, args []string) error {
 
 	env := environ(os.Environ())
 	secretStore := getSecretStore()
+	envVarKeys := make([]string, 0)
 	for _, service := range services {
 		if err := validateService(service); err != nil {
 			return errors.Wrap(err, "Failed to validate service")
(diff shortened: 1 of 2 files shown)


systemizer:
hey Yarek! Thanks for submitting this pull request.

This looks useful. What do you think about adding instead a global debug flag `-d` (instead of a subcommand flag `-k`)? This will support adding debug information for other commands in the future as well. 

An example of how this is done can be found in our aws-okta library: https://github.com/segmentio/aws-okta/blob/master/cmd/root.go#L26

cmd/exec.go
```diff
@@ -52,13 +58,19 @@ func execRun(cmd *cobra.Command, args []string) error {
...
+
 			if env.IsSet(envVarKey) {
 				fmt.Fprintf(os.Stderr, "warning: overwriting environment variable %s\n", envVarKey)
 			}
 			env.Set(envVarKey, rawSecret.Value)
 		}
 	}
 
+	if printKeys {
+		fmt.Fprintf(os.Stderr, "info: With environment %s\n", strings.Join(envVarKeys, ","))
```
systemizer: let's print to stdout instead of stderr

YarekTyshchenko:
Called it verbose in hope that it doesn't encourage exposure of any of the secrets. Thoughts @systemizer?

PR #134, opened 2018-09-13
Fix flag shorthand conflict.

jmcarp:
Both the "verbose" flag in the cli root and the "version" flag in the
read command use the "v" shorthand, which causes a panic on read. This
patch drops the shorthand for the "version" flag.

@nickatsegment


nickatsegment:
@jmcarp Thanks for this! Panics: they're bad.

My gut feeling is that we should drop the short form of verbose because I think it's newest, and isn't yet in a release. Will take a closer look tomorrow.

jmcarp:
I didn't realize the `--verbose` flag was newer. Just amended to drop the shorthand for `--verbose` instead of `--version` to avoid a breaking change.

PR #136, opened 2018-09-19
Adding analytics for usage metrics

systemizer:
This adds metrics to chamber so we can track internal company usage of the tool. Chamber built for open source will not have the linked analytics key.

This PR basically copies the way metrics were added to aws-okta

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	analytics "github.com/segmentio/analytics-go"
 	"github.com/segmentio/chamber/store"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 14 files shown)


cmd/delete.go
```diff
@@ -4,6 +4,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	analytics "github.com/segmentio/analytics-go"
```
f2prateek: I recommend using https://github.com/segmentio/analytics-go/tree/v3.0. We've already deprecated v2 and are just waiting for until the deprecation date to merge v3.0 into master.

systemizer: yup! I'm using v3 :)

f2prateek: ah 👍

PR #138, opened 2018-09-21
add circle jobs to test installation with `go get`

nickatsegment:


eculver:
LGTM, but it looks like this should depend on #123 otherwise this will generate new `go.{mod,sum}` every time its run w/ `GO111MODULE=on`.

nickatsegment:
Superseded by #144

Issue #128, opened 2018-09-05
feature request: Having a namespace in front of the ssm /service/ path. 


maartenvanderhoef:
Hi there!

What are the thoughts of having an ENV VAR `SSM_PATH_NAMESPACE` or `SSM_PATH_PREPEND` which prepends a string in front of the `/service/lookupvar` . This to make it easier to organize SSM a bit when it's not only docker applications looking up SSM. For example I'm currently using `/applications/service/lookupvar`.

emmanuel:
Just trying out Chamber for the first time after reading the post on the AWS blog. Thanks for the tool (and the article), you helped me quickly get up to speed with effective patterns for using Parameter Store. 

However, I support the request in this issue. Namely: Chamber doesn't support using path-based organization of my parameters, which I take to be recommended practice (based on the Parameter Store docs and examples I've seen on AWS).

In the meantime, I've got a systemd unit that does the basic version. In other words, the below fetches all parameters within a path and upcases the keys then outputs an environment file (in the format "ENV_VAR=decrypted_value"):

```
[Unit]
Description=Fetch app startup config from AWS Parameter Store
Requires=pip-install.service
After=pip-install.service

[Service]
Type=oneshot
Environment="RAW_PARAMETER_OUTPUT=/tmp/parameter_output.json"
Environment="PARAMETER_ENV_FILE=/run/parameter_store.env"
Environment="AWS_DEFAULT_REGION=${AWS_REGION}"
ExecStartPre=/bin/bash -c 'aws ssm get-parameters-by-path --path="${PARAMETER_STORE_PATH}" --with-decryption > $${RAW_PARAMETER_OUTPUT}'
ExecStart=/bin/bash -c 'jq -r \'.Parameters | map((.Name | split("/")[-1] | ascii_upcase) +"="+ (.Value))[]\' $${RAW_PARAMETER_OUTPUT} > $${PARAMETER_ENV_FILE}'
ExecStartPost=/usr/bin/shred -u $${RAW_PARAMETER_OUTPUT}
```

AndreyMarchuk:
+1 to prefix (namespace) or at least allow '/' in service name. Currently using '/' gives an error:

> Error: Failed to validate service: Failed to validate service name 'namespace/env/service'.  Only alphanumeric, dashes, and underscores are allowed for service names

AndreyMarchuk:
This might be resolved by https://github.com/segmentio/chamber/pull/118

[...]

PR #140, opened 2018-09-24
Allow prefix to be set that is prepended to services

darend:
## what

When `CHAMBER_SERVICE_PREFIX` is set, the value is prepended to all services when access the store

## why

Issue: https://github.com/segmentio/chamber/issues/128

This simplifies organizing keys in the store. A common use case would when you have multiple environments that the services are deployed to. By setting the prefix the same chamber command be run across all environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,20 @@ for more details.
 
 If you'd like to use a different region for chamber without changing `AWS_REGION`, you can use `CHAMBER_AWS_REGION` to override just for chamber.
 
+### Service Prefix
+
+If `CHAMBER_SERVICE_PREFIX` is set, chamber will prefix all services with its value. This can be useful when you want to namespace your configuration without having to repeat it in each service key.
+
+For example, these are equivalant:
+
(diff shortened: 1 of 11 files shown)


osterman:
@nickatsegment or @systemizer do you have any reservations about this PR? Seems pretty popular.

README.md
```diff
@@ -206,6 +206,20 @@ for more details.
...
 
+### Service Prefix
+
+If `CHAMBER_SERVICE_PREFIX` is set, chamber will prefix all services with its value. This can be useful when you want to namespace your configuration without having to repeat it in each service key.
+
+For example, these are equivalant:
+
+```
+# Store value under 'applications\service\key'
+chamber write applications\service key value
```
nickatsegment: Surely you mean `applications/service/key`?

nickatsegment:
Closing this. If people want this, I'd like to see some evidence that the same thing's not accomplishable with a bit of `sh`

Issue #147, opened 2018-09-27
Should chamber vendor its dependencies in `vendor/`?


nickatsegment:
We have `vendor/vendor.json` listing the packages, but we don't have them checked into source.

nickatsegment:
@dfuentes Do you know why this is?

nickatsegment:
Yeah, I agree: `go get` has too many unknowns to be supported as the main installation mechanism. A while back I had updated the README with a link to the wiki to install. Maybe the README should just say to download a binary as the main installation method, packages 2nd, and `go get` last.

nickatsegment:
Even if we only support go modules, this is still kind of an open question to me. My heart says "you should be able to compile chamber offline with only the git repo and a go compiler, if you really have to". But of course, continuously testing that that works annoying and dull and perhaps a pointless chore. Also, I've seen conflicts when using other libraries that have vendored dependencies.

One thing's clear: we should either support it, or it should be empty (save vendor.json for go1.10-ers)

PR #148, opened 2018-09-27
Add packages to source control to deal with issues with analytics-go

systemizer:


nickatsegment:
Fixes https://github.com/segmentio/chamber/issues/147

nickatsegment:
@brycehendrix Yep, we use `govendor` and new native Go modules. Unfortunately `go get` doesn't always behave the way we want. This is specifically in response to a problem with github.com/segmentio/analytics-go using git submodules to pull in its "vendored" dependencies, but that issue should be fixed https://github.com/segmentio/analytics-go/pull/136 

(@systemizer maybe that means this PR isn't needed?)

See also
https://github.com/segmentio/chamber/issues/147
https://github.com/segmentio/chamber/wiki/Go-Modules

systemizer:
Cool, didn't know it was fixed. Yeah, we can get rid of this. Will open a PR to revert the revert!

Issue #143, opened 2018-09-25
Feature request: Having a simple way to disable chamber backend lookups


maartenvanderhoef:
Hi everyone,

In some use cases the transformation into using Chamber inside an (external) organisation is not always straight forward. In many cases it is not only a migration on AWS from using ENV to using SSM/KMS, but also a migration from other providers like Heroku to AWS.

Still to be able to get prepared with applications still sitting outside of AWS with various teams and docker orchestrations, it would be great to already have chamber implemented in a lazy way. As it's currently implemented one would have to change the `ENTRYPOINT` or `CMD` and this is just not always possible.

I suggest to be able to switch of any backend lookups by having a None backend which can be used by having an the env variable `CHAMBER_SECRET_BACKEND` to be `NONE` The logic to have a different backend is already there and the PR https://github.com/segmentio/chamber/pull/127 is the most simple implementation to achieve exactly this. Happy to hear your comments, and good work everybody!

PR #149, opened 2018-09-27
Addition of a NULL Backend to provide a simple way of disabling backend lookups

maartenvanderhoef:
## What

The addition of a `NULL` backend store, next to `s3` and `ssm`. Invoked by the ENV variable `CHAMBER_SECRET_BACKEND` set to `NULL`. The null-backend has no actual backend. Chamber will simply transparently pass through ENV vars.  https://github.com/segmentio/chamber/issues/143 is the related issue.

## Why

In some use cases the transformation into using Chamber inside an (external) organisation is not always straight forward. In many cases it is not only a migration on AWS from using ENV to using SSM/KMS, but also a migration from other providers like Heroku to AWS. To be able to get an organisation prepared with applications still sitting outside of AWS, with various teams and various docker orchestrations, it would be great to already have chamber implemented in a lazy way without the need to set the`ENTRYPOINT` or `CMD`, as this is not always possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,6 +214,13 @@ To configure chamber to use the S3 backend, set `CHAMBER_SECRET_BACKEND` to `S3`
 
 This feature is experimental, and not currently meant for production work.
 
+## Null Backend (experimental)
+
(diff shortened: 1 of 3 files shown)


systemizer:
merged! thanks again @maartenvanderhoef !

PR #150, opened 2018-09-28
Quote environment variables.

jmcarp:
Fixes dotenv export for variables with newlines and other special
characters.

Escape helper borrowed from https://github.com/joho/godotenv/blob/master/godotenv.go (MIT licensed).

### DIFF
--- a/cmd/export.go
+++ b/cmd/export.go
@@ -15,6 +15,8 @@ import (
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


jmcarp:
cc @nickatsegment @systemizer

nickatsegment:
Looks like a good idea to me. It'd be cool if you could accompany this with tests showing a few pathological examples and how this fixes them.

jmcarp:
Added tests.

PR #153, opened 2018-10-07
Fix typo in export example

aandrieu:
The json in example for `export` command is not valid. Fix the example output.


nickatsegment:
Sorry for the delay, and thanks

Issue #86, opened 2018-04-25
Reading versions of secrets greater than v50 but less than current version fails


amasare:
A `chamber read my_service my_secret -v 56` of a secret named `my_secret` with current version as version 56 will give the current secret, but a `chamber read my_service my_secret -v 50<version_number<current` of any versions of the secret between v51 and current returns `Error: Failed to read: secret not found`.

For eg `chamber read my_service my_secret -v 55` or `chamber read my_service my_secret -v 54` or `chamber read my_service my_secret -v 53` or `chamber read my_service my_secret -v 52` or or `chamber read my_service my_secret -v 51` fails.

A `chamber read my_service my_secret -v 50` and version numbers below 50 returns values correctly.

mechanical-fish:
I can replicate this entertaining bug. Will spend a few minutes trying to learn Go debugging well enough to figure this one out!

mechanical-fish:
Found it; it's pretty simple: The AWS SSM `GetParamaterHistory` API call has a built-in limit of 50 return values at a time. This code needs to be fixed to use `GetParameterHistoryPages` instead.

PR #158, opened 2018-10-13
Fix missing history and failed reads beyond secret version 50

mechanical-fish:
Fixes issue #86 - because the `GetParameterHistory` API is limited to 50 results per request, secret versions greater than 50 did not show up in the history and could not be retrieved.


nickatsegment:
Love it. Sorry for the delay.

PR #162, opened 2018-11-01
Support case sensitivity for a few commands

### LINKED ISSUE #160: Import should lowercase all keys
When importing a JSON file with uppercase keys, they are imported as is. If I write a key via `chamber write`, the key is lowercased. This leads to duplicated keys. Plus, it breaks the read / delete command, since it apparently assumes lowercased keys.

acaire:
This will add a `--case-sensitive` switch to the following commands:

    * delete
    * history
    * import
    * read

Notably changing the import behaviour to fail an import payload with
mixed case keys

This follows on from previous discussions:

1. #160 is the latest outstanding issue with no comments yet
2. #21 was closed citing duplication protection from user error
3. #66 was closed citing #21 and consistency concerns

In detail this commit will:

  1. Modify `import` to fail an import payload when mixed case keys are provided except when the `--case-sensitive` switch is passed - This closes #160 with the following considerations:
    a. This protects users from themselves
    b. This helps anyone using chamber that currently bulk imports a mismash of case in a single export to be notified to correct their behaviour
    c. This new default behaviour mimics the manual write call that has been defended in similiar PRs, but still allows for the old behaviour with a switch
  2. Modify `delete` by adding a non default `--case-sensitive` switch to delete case sensitive params (which helps clean up any messes made by an existing import)
  3. Modify `history` by adding a non default `--case-sensitive` switch to allow individual case sensitive param history to be read (which allows viewing of history for keys already seen with `list`)
  4. Modify `read` by adding a non default `--case-sensitive` switch to allow individual case sensitive params to be read (which has a tremendous amount of value in allowing a user to read any existing key that wasn't imported with Chamber)

Here's a summary of the current core chamber functions:

  #### Four commands with enforced lowercase consistency
  * `delete` will lowercase all delete calls to SSM
  * `history` will lowercase all history calls to SSM
  * `read` will lowercase all reads from SSM
  * `write` will lowercase all puts to SSM

  #### Four commands break this consistency
  * `exec` returns values as is from SSM except for the `dotenv` output, which will uppercase all keys and keep duplicates
  * `export` will respect case sensitivity, and fetch case sensitive params as is in SSM except the resultant value will be unpredictable, but at least there's a warning about it
  * `import` will respect case sensitivity, and store case sensitive params as is in SSM
  * `list` will respect case sensitivity, and list case sensitive params as is in SSM

Finally, this commit intends to continue to protect the user from doing the wrong thing when writing values thus improving the end user experience, whilst reducing the barrier for new users of Chamber by supporting an existing key structure without compromising on new parameter ideologies

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -18,16 +18,23 @@ var deleteCmd = &cobra.Command{
 }
 
 func init() {
+	deleteCmd.Flags().BoolVarP(&caseSensitive, "case-sensitive", "", false, "Enable case sensitive service names and keys. Defaults to lowercase keys and services")
 	RootCmd.AddCommand(deleteCmd)
 }
 
 func delete(cmd *cobra.Command, args []string) error {
 	service := strings.ToLower(args[0])
+	if caseSensitive {
+		service = args[0]
+	}
 	if err := validateService(service); err != nil {
 		return errors.Wrap(err, "Failed to validate service")
 	}
 
 	key := strings.ToLower(args[1])
+	if caseSensitive {
+		key = args[1]
+	}
 	if err := validateKey(key); err != nil {
 		return errors.Wrap(err, "Failed to validate key")
 	}
--- a/cmd/history.go
+++ b/cmd/history.go
@@ -21,16 +21,23 @@ var historyCmd = &cobra.Command{
 }
 
 func init() {
+	historyCmd.Flags().BoolVarP(&caseSensitive, "case-sensitive", "", false, "Enable case sensitive service names and keys. Defaults to lowercase keys and services")
 	RootCmd.AddCommand(historyCmd)
 }
 
 func history(cmd *cobra.Command, args []string) error {
 	service := strings.ToLower(args[0])
+	if caseSensitive {
+		service = args[0]
+	}
 	if err := validateService(service); err != nil {
 		return errors.Wrap(err, "Failed to validate service")
 	}
 
 	key := strings.ToLower(args[1])
+	if caseSensitive {
+		key = args[1]
+	}
 	if err := validateKey(key); err != nil {
 		return errors.Wrap(err, "Failed to validate key")
 	}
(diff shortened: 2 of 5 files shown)


acaire:
@dfuentes I'm hoping this will address some of the concerns made in previous attempts

nickatsegment:
Thanks @acaire 	, love the write-up. Two things I would love to see but wouldn't keep me from merging this PR:
- tests showing the behaviours you're talking about
- a new section in the README about case sensitivity gotchas; this probably should have been done long ago

osterman:
> Finally, this commit intends to continue to protect the user from doing the wrong thing when writing values thus improving the end user experience

Just gonna throw this use-case out there... the case squashing is not appreciated by us terraform users. Terraform by convention uses lower case variable names. To set these as environment variables, we prefix them by [`TF_VAR_`](https://www.terraform.io/docs/configuration/variables.html#environment-variables). Yes, we can uppercase our variables in HCL and it "works" (and [we have been](https://github.com/cloudposse/terraform-root-modules/blob/master/aws/backing-services/aurora-mysql.tf)), but I don't like that `chamber` is opinionated down to this degree and especially so that it forces us to break convention in another context. It also makes our modules less portable across organizations that might not use `chamber` for these settings and now need to define upper and lower case envs for the same settings. 

Anyways, we wrote [`tfenv`](https://github.com/cloudposse/tfenv) as a temporary hack around this inconvenience. My vote is for `chamber` to respect `--case-sensitive` across the board.

af6140:
Agree with @osterman , we use aws-env to export env variables for ssm parameters, forcing lower case is not preferred behavior.

aharasta-earnup:
@nickatsegment Is there anything I can do to help get this PR merged?

nickatsegment:
@aharasta-earnup 	Yep, I was looking for some more tests and documentation in https://github.com/segmentio/chamber/pull/162#issuecomment-435521016

nickatsegment:
I don't think this is getting done without breaking API changes, so I've targeted for v3.0

PR #167, opened 2018-11-07
add NewS3StoreWithBucket constructor

nickatsegment:
Deprecate NewS3Store to avoid env var in lib

### DIFF
--- a/store/s3store.go
+++ b/store/s3store.go
@@ -18,7 +18,8 @@ import (
 
 const (
 	MaximumVersions = 100
-	BucketEnvVar    = "CHAMBER_S3_BUCKET"
+	// deprecated
+	BucketEnvVar = "CHAMBER_S3_BUCKET"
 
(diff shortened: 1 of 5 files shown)


store/s3store.go
```diff
@@ -56,7 +57,17 @@ type S3Store struct {
 	bucket string
 }
 
+// Deprecated
```
systemizer: maybe add to this comment what to use instead

systemizer:
the one thing i worry about is whether someone is depending on the environment variable with the library. But given this is an experimental feature and we haven't started using it, I agree with this approach

nickatsegment:
I think the reliance on an env var for a critical piece of info makes its utility as a library pretty limited. You'd need a lock around creation to make sure the env var is in the correct state.

I considered just replacing NewS3Store. But I think deprecating and targeting removal in v3 is the best way.

PR #169, opened 2018-11-14
export environ as its own lib

nickatsegment:
I could have just copy-pasted this stuff, but there's a fair amount of fiddly details that I think could use capturing in their own exported library

This is for another docker entrypoint system that I'm working on that would work better if chamber weren't an extra binary.

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/pkg/errors"
+	"github.com/segmentio/chamber/environ"
 	"github.com/spf13/cobra"
 	analytics "gopkg.in/segmentio/analytics-go.v3"
 )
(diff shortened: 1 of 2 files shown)


environ/environ.go
```diff
@@ -0,0 +1,85 @@
...
+// Load loads environment variables into e from s given a service
+// collisions will be populated with any keys that get overwritten
+func (e *Environ) Load(s store.Store, service string, collisions *[]string) error {
+	return e.load(s, service, collisions, false)
+}
+
+// LoadNoPaths is identical to Load, but uses v1-style "."-separated paths
+//
+// Deprecated like all noPaths functionality
+func (e *Environ) LoadNoPaths(s store.Store, service string, collisions *[]string) error {
```
nickatsegment: This split is a little non-intuitive. Eventually we'll remove the NoPaths logic, and I'd like to keep the Load function. At that point we can roll `key` and `load` into `Load`.

cmd/exec.go
```diff
@@ -50,67 +51,35 @@ func execRun(cmd *cobra.Command, args []string) error {
...
-		rawSecrets, err := secretStore.ListRaw(strings.ToLower(service))
+		collisions := make([]string, 0)
+		var err error
+		if _, noPaths := os.LookupEnv("CHAMBER_NO_PATHS"); noPaths {
+			err = env.LoadNoPaths(secretStore, service, &collisions)
+		} else {
+			err = env.Load(secretStore, service, &collisions)
+		}
 		if err != nil {
 			return errors.Wrap(err, "Failed to list store contents")
```
Fauzyy: Can we just return `err` here? Since we already return `errors.Wrap(err, "Failed to list store contents")` from `Load` and `LoadNoPaths`

nickatsegment: Yep, good point. Updated to return err directly from `Load*`

nickatsegment:
@systemizer 	Yes, that's a possibility. My stance is that as soon as you export a symbol, you should assume it's widely used and should endeavour to not break contract.

nickatsegment:
Realizing that I should have made Load() accept a new interface that only requires ListRaw, since that's all it really needs. Would make mocking simpler.

PR #175, opened 2018-11-27
add write --skip-unchanged flag

q3aiml:
By default a write attempt results in a new history entry even when the value is unchanged. Under most use this is no big deal but I've run into cases where we update secrets through automation, for example syncing during transitions between systems, and the history is filled with a bunch of non-changes. This adds a simple option to ignore writing the secret if the value is unchanged.

This is also easy to accomplish with a tiny bit of scripting but it has come up in a few different cases for me so I figure I would see if there is any interest in baking it in. Putting it in the tool does give the option to later push the implementation down into the store implementations if someone wanted to save a double read with the s3 backend, for example.

Thanks for looking!

### DIFF
--- a/cmd/write.go
+++ b/cmd/write.go
@@ -13,7 +13,8 @@ import (
 )
 
 var (
-	singleline bool
+	singleline    bool
+	skipUnchanged bool
 
 	// writeCmd represents the write command
 	writeCmd = &cobra.Command{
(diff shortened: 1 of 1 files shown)


nickatsegment:
For me, I'd say having this on by default would be best, with a flag to force a write. But I'm not sure that's the general opinion, so let's just get this in.

q3aiml:
@nickatsegment thanks! I agree that it seems to me to make sense for this to be the default. In addition to not knowing what others think, I also wonder if anyone would get surprised by write newly requiring SSM read and key decrypt perms, or if it deserves a bigger version bump

PR #179, opened 2019-01-15
update AWS SDK to v1.15.90

q3aiml:
Picking up support for credential_source. This allows assuming a role
while using profile creds from ec2 instance metadata:
https://github.com/aws/aws-sdk-go/pull/2201.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,10 @@
 module github.com/segmentio/chamber
 
 require (
-	github.com/aws/aws-sdk-go v1.15.46
+	github.com/aws/aws-sdk-go v1.15.90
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,8 +1,10 @@
 module github.com/segmentio/chamber
 
 require (
-	github.com/aws/aws-sdk-go v1.15.46
+	github.com/aws/aws-sdk-go v1.15.90
 	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/go-ini/ini v1.25.4 // indirect
```
nickatsegment: It's kind of weird that a patch release would add dependencies. But meh

PR #182, opened 2019-01-31
add global --backend flag

nickatsegment:
`--backend` takes an option that is either a bare backend name like `ssm` or a URL like `s3://<bucket>`. I can imagine a world where it could take query options like `ssm://?use-v1-paths=1` instead of passing those by yet another env var*

This also extends the current `CHAMBER_SECRET_BACKEND` to allow for the URL syntax as well.

Besides that one, this gets us away from the proliferation of gross env vars. IMO env vars are just a bad UX vs flags. For one, none of our env vars are documented in the online help, only the README. This flag is documented in `--help`.

* This would require detangling the scattering of env vars all over the lib code base. I started trying to do that, but not sure it's worth the effort https://github.com/segmentio/chamber/pull/183

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,10 +2,12 @@ package cmd
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"regexp"
 	"strings"
 
+	"github.com/pkg/errors"
 	"github.com/segmentio/chamber/store"
 	"github.com/spf13/cobra"
 	analytics "gopkg.in/segmentio/analytics-go.v3"
@@ -20,7 +22,9 @@ var (
 	verbose        bool
 	numRetries     int
 	chamberVersion string
-	backend        string
+	// one of *Backend consts
+	backend     string
+	backendFlag string
 
 	analyticsEnabled  bool
 	analyticsWriteKey string
@@ -42,7 +46,8 @@ const (
 	S3Backend   = "S3"
 
 	BackendEnvVar = "CHAMBER_SECRET_BACKEND"
-	BucketEnvVar  = "CHAMBER_S3_BUCKET"
+	// deprecated
+	BucketEnvVar = "CHAMBER_S3_BUCKET"
 )
 
 var Backends = []string{SSMBackend, S3Backend, NullBackend}
@@ -59,6 +64,12 @@ var RootCmd = &cobra.Command{
 func init() {
 	RootCmd.PersistentFlags().IntVarP(&numRetries, "retries", "r", DefaultNumRetries, "For SSM, the number of retries we'll make before giving up")
 	RootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "", false, "Print more information to STDOUT")
+	RootCmd.PersistentFlags().StringVarP(&backendFlag, "backend", "b", "ssm",
+		`Backend to use; AKA $CHAMBER_SECRET_BACKEND
+	null: no-op
+	ssm: SSM Parameter Store
+	s3://<bucket>: S3, using <bucket>`,
+	)
 }
 
 // Execute adds all child commands to the root command sets flags appropriately.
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -59,6 +64,12 @@ var RootCmd = &cobra.Command{
 func init() {
 	RootCmd.PersistentFlags().IntVarP(&numRetries, "retries", "r", DefaultNumRetries, "For SSM, the number of retries we'll make before giving up")
 	RootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "", false, "Print more information to STDOUT")
+	RootCmd.PersistentFlags().StringVarP(&backendFlag, "backend", "b", "ssm",
```
systemizer: seems strange that this value is either an "enum"-like type or an s3 url. how about the valid values are "ssm", "s3", or "null" and then if it is "s3", we require another flag called `--s3-backend-bucket`?

systemizer: hmm yeah i didn't see ssm:// as being a meaningful URL.

i agree requiring both env var and flags will be confusing. Requiring flags for a backend configuration is annoying, but we're already requiring the user to specify a backend override via flags to use s3.

could we recommend users to use aliases for default backends? 
`alias chambers3=chamber --backend s3 --s3-backend-bucket mybucket`

we already recommend this pattern with other use cases (e.g. aws-okta):
`alias chamberprod=aws-okta exec production -- chamber`

boggsboggs: I would find `--backend s3://mysecrets` easiest to use. Seems like it makes the code inside chamber uglier though. Given the choice between easy for the user and easy for the developer, with chamber, I'd choose easy for the user. It's a relatively small/simple codebase where usability matters a lot.

Recommending some aliases seems like a solid way to improve usability:
```
alias chambers3prod=aws-okta exec prod-privileged -- chamber --backend s3 --s3-backend-bucket prodbucket
alias chambers3stage=aws-okta exec stage-privileged -- chamber --backend s3 --s3-backend-bucket stagebucket
```

boggsboggs: In terms of env vars vs. command line, I usually assume that command line takes precedence, since it's more tightly scoped

systemizer: here's my stance

1. creating aliases is very easy and fixes any UX issues someone might have by required flags
2. I am interpreting the current backend config  as a string (and not a url). i would expect users to think the same if the values were currently "null" and "ssm"
3. given the current config is not explicitly a url, setting it as a url could cause confusion for a future engineer. if we add a new backend in the future that requires multiple configuration options, we're going to need to add flags (unless we want to somehow overload this setting with more options via some string formatting)

this case isn't that big of an issue. but it might set the standard for future backends. If I were adding this, I'd personally add the additional command line flag, but I won't fight for it if you feel strongly

nickatsegment: > if we add a new backend in the future that requires multiple configuration options, we're going to need to add flags (unless we want to somehow overload this setting with more options via some string formatting)

I was thinking that if we need options, they'd be query params, like `s3://bucket?retries=5`.

Personally, I like the url syntax because it's so compact vs the extra flags (especially if we grow more flags). Maybe I'm used to it from other apps like `gsutil` and DB connection strings, but I can imagine that being unfamiliar and weird.

Anyway, besides s3, there's really no use case for it at present, so I think I'll just back that out in favour of specific flags. If we get more backends or more options, we can revisit then.

How do y'all feel about `--backend-<backend>-<option>`? i.e. `--backend-s3-bucket` Just so they sort together.

boggsboggs: `--backend-s3-bucket` sounds good to me

nickatsegment: Ok, I've removed the URI stuff in the latest.

cmd/root.go
```diff
@@ -59,6 +63,13 @@ var RootCmd = &cobra.Command{
...
+		`Backend to use; AKA $CHAMBER_SECRET_BACKEND
+	null: no-op
+	ssm: SSM Parameter Store
+	s3: S3; requires --backend-s3-bucket`,
+	)
+	RootCmd.PersistentFlags().StringVarP(&backendS3BucketFlag, "backend-s3-bucket", "", "", "bucket for S3 backend; AKA $CHAMBER_S3_BUCKET")
```
nickatsegment: One thing that I don't like is that the user can do something invalid like `chamber -b s3 --backend-s3-bucket=mybucket` and it's not an error. Validating that only the correct flags are set in the correct backend modes sounds like a chore, especially when you factor in the env vars. Well, it's probably not so bad when we only have one backend-specific flag.

nickatsegment:
Ok, ready for another review pass @boggsboggs 	@systemizer

PR #184, opened 2019-02-02
Add env command to export secrets

### LINKED ISSUE #94: Feature Request: Output shell script to set environment variables
Like using `exec` for a specific process, something like `chamber populate` will grab secrets and then evaluate them or `export` into the current environment.

jradtilbrook:
This resolves #94 by adding a new command that will print the secrets in a format that can be `eval`'ed to export them as environment variables.

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -1,0 +1,76 @@
+package cmd
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


osterman:
Thank you!!! 

We can finally stop doing:
```
source <(chamber exec service -- sh -c 'export -p')
```

PR #186, opened 2019-02-07
Dockerfile

nickatsegment:
This doesn't build/publish automatically yet, but it's a start. Enables you to do

```
FROM segment/chamber:2 AS chamber
FROM scratch

COPY --from=chamber /chamber /bin/chamber

COPY ./myapp /bin/myapp

ENTRYPOINT ["/bin/chamber", "exec", "myapp", "--", "/bin/myapp"]
```


nickatsegment:
@systemizer 	It's the next Kubernetes. So hot right now.

Issue #141, opened 2018-09-24
Have a mechanism for finding services


majormoses:
Currently you can list all the secrets for a service but there is no way to find what the service is called. I think there should be some command to help with this.

osterman:
Agree, this would be helpful.

nickatsegment:
I'm not sure off the top of my head whether we add any metadata to SSM to indicate that it's a chamber secret.

Could you not just use the AWS CLI? Eg. `aws ssm get-parameters-by-path --path /`

majormoses:
> Could you not just use the AWS CLI? Eg. `aws ssm get-parameters-by-path --path /`

We have tens of thousands of parameters, someone on our team tested this (using `--recursive`) in one of our accounts and it took something like a full day to complete. His exact command was:
```
aws-vault exec MY_ACCOUNT -- aws ssm get-parameters-by-path --path / --recursive  | jq  '.Parameters|.[]|.Name' | awk -F\" {'print $2'} 
```

nickatsegment:
Hm, that's strange: I listed our secrets (2.5k) in about 10 seconds...

Chamber would have to use the same API call, so it wouldn't be any faster.

nickatsegment:
Yeah, that makes sense. I'm sure we've had to have our limits increased at some point.

nickatsegment:
With the SSM backend at least, there's no way I can think of that would be better in terms of UX or efficiency than just using the AWS CLI. Counterpoints appreciated.

majormoses:
> With the SSM backend at least, there's no way I can think of that would be better in terms of UX or efficiency than just using the AWS CLI. Counterpoints appreciated.

I agree there is probably not anything from a performance perspective but needing to memorize/type/alias but from a UX perspective this is pretty far from ideal:
```
aws ssm get-parameters-by-path --path / --recursive  | jq  '.Parameters|.[]|.Name' | awk -F\" {'print $2'} 
```
Would like there to be something much simpler such as `chamber list-secrets` even if all it did was execute that command on the users behalf.

nickatsegment:
Yeah, you're right. I can see value in that. Especially if you were working with multiple backends or the v1/v2 migration. 

I'd say `list-services` would be a slightly better name to keep with the metaphors of services containing many secrets.

We'd need to make sure each backend is supported, and probably add a new func to the `store` interface, say `ListServices`.

I've added the `help wanted` label, since I don't think the maintainers at Segment have an immediate need for this.

[...]

PR #187, opened 2019-02-13
Add search command for SSM backend

pecigonzalo:
This is a draft implementation of a search command.

In many cases we need to search for services and we have to fallback to `aws ssm` which is ok,
but it could be a good idea to implement this simple search functionality in `chamber`.

Let me know your thoughts and ill add the missing parts

### Example output
Service name: `servicename`
```
❯ awsve dev.admin -- dist/chamber-v2.3.3-pre2-5-g059f1d1-dev-linux-amd64 search /servicena  
Service
/servicename/secreta
/servicename/secretb
```

#### TODO

- [ ] Function/Method documentation
- [ ] Any required arguments?
- [ ] Implement S3
- [ ] Tests

### DIFF
--- a/cmd/list-services.go
+++ b/cmd/list-services.go
@@ -1,0 +1,60 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"sort"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+// listServicesCmd represents the list command
+var listServicesCmd = &cobra.Command{
+	Use:   "list-services <service>",
+	Short: "List services",
+	RunE:  listServices,
+}
+
+var (
+	includeSecretName bool
+)
+
+func init() {
+	listServicesCmd.Flags().BoolVarP(&includeSecretName, "secrets", "s", false, "Include secret names in the list")
+	RootCmd.AddCommand(listServicesCmd)
+}
+
+func listServices(cmd *cobra.Command, args []string) error {
+	var service string
+	if len(args) == 0 {
+		service = ""
+	} else {
+		service = strings.ToLower(args[0])
+
+	}
+	secretStore, err := getSecretStore()
+	if err != nil {
+		return errors.Wrap(err, "Failed to get secret store")
+	}
+	secrets, err := secretStore.ListServices(service, includeSecretName)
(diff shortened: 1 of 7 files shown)


pecigonzalo:
@nickatsegment sorry to tag you directly but, could you confirm if this is desired? so I can finish the TODO list.

nickatsegment:
@pecigonzalo 	No problem at all. Feel free to do so.

A few design questions:
- Are we searching for services or secrets? Your example shows a service name as input, but lists all secrets for that service as output.
- Define "search" :) Looks like from your code it's just looking for a prefix. What about partial matches? We need to make sure it's simple to implement for all backends. (The S3 backend could be pretty dynamic since it's got an index file, but we need to cater to the lowest common subset.)

Also, this sounds very close to https://github.com/segmentio/chamber/issues/141. Maybe instead of calling it "search", it could look like `chamber list-services --prefix=servicena` or `chamber list-services 'servicena*'`

pecigonzalo:
@nickatsegment

It would search for a "path", the same thing can be implemented in S3 I believe. Im building on SSM "limitations" as that is the original store, and the API is a bit more limited, I believe it does not support searching for partials AFAIK, but ill try and if possible implement it. We would have to list the entire SSM store and filter client side, but that would be quite ugly and inefficient. 

I was thinking more in the lines of `beginsWith` and just list all "childs", even the secret name, but I can strip that if desired, I thought it might be useful.
Maybe we can just support a couple of search options. 

Regarding #141, yeah, i think it will "fix" that issue.

Given there is an appetite for it, ill start doing a proper implementation supporting this.

pecigonzalo:
Just FYI, this is still in my radar, I just did not have the time to finish it.

pecigonzalo:
@nickatsegment renamed and implemented a bit better for SSM, tests still missing.
I was trying to implement the S3 one, but given the current store, implementation is experimental and makes a bit hard to look for secrets. 
I can do a `ListObjects` recursively and filter the `__latest.json` files from the output. What do you think?


For the time being, I implemented SSM and put a notice for S3, in case you want to merge this with SSM only.

nickatsegment:
Thanks for following through @pecigonzalo !

osterman:
Thanks @pecigonzalo !! Excited for this enhancement.

PR #195, opened 2019-03-12
feat: New S3 Backend for using KMS Keys as S3 Encryption Method

matthewborden:
Previously we were using AWS Parameter Store. However after hitting AWS Rate Limiting on Parameter Store and being refused a limit increase we decided to switch the Chamber S3 Backend. However, a major point of difference between the SSM Backend and the S3 Backend is that the SSM Backend supports AWS KMS Keys.

This PR switches from using AES256 Encryption (read the one key stored by all S3 servers) to using KMS Keys for Encryption (read specific keys that you create via AWS KMS Service, AWS Manages the KMS key but you have control over disabling them when they become compromised and restricting access to them via IAM Roles).

Along the way, I ran into the issue that Chamber is caching the keys and values for each service in a __latest.json file. To make the KMS Encryption worthwhile, I changed the __latest.json be a "latest" file per KMS Key. This differs slightly from the behavior of the SSM Store backend when the user doesn't provide the CHAMBER_KMS_KEY_ALIAS environment variable. In the SSM backend, you as the user are able to read any secrets which you have the KMS Key for. However In the S3 Version (with the changes that I've made, you're not able to see secrets encrypted with a KMS Key that you do not have access to).

Currently, the S3 Bucket looks like:

- service-name:
   - __latest.json
   - Key Named JSON File (eg, database_name.json)
   - Key Named JSON File (eg, database_password.json)
   - Key Named JSON File (eg, database_username.json)
- service-name:
   - __latest.json
   - Key Named JSON File (eg, database_name.json)
   - Key Named JSON File (eg, database_password.json)
   - Key Named JSON File (eg, database_username.json)

With these changes:

- service-name:
  - alias_default_parameter_store_key__latest.json
   - Key Named JSON File (eg, database_name.json)
   - Key Named JSON File (eg, database_password.json)
   - Key Named JSON File (eg, database_username.json)
- service-name:
   - alias_default_parameter_store_key__latest.json
   - alias_very_secret_kms_key__latest.json
   - Key Named JSON File (eg, database_name.json)
   - Key Named JSON File (eg, database_password.json)
   - Key Named JSON File (eg, database_username.json)

Implementation Decisions:
- I've chosen to make `chamber list` not show secrets that are encrypted with KMS Keys that you do not have access to. This is because multiple KMS Keys can be used within a single chamber service and some readers will only have one key.
- `chamber write` and `chamber delete` requires CHAMBER_KMS_KEY_ALIAS to be set to the same KMS Key as an existing Chamber secret. This avoids the latest version of a secret being left encrypted in S3 after you move the secret to a new KMS key.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,29 @@ To configure chamber to use the S3 backend, use `chamber -b s3 --backend-s3-buck
 
 This feature is experimental, and not currently meant for production work.
 
+### S3 Backend using KMS Key Encryption (Experimental)
+
+This backend is similar to the S3 Backend but uses KMS Key Encryption to encrypt your documents at rest, similar to the SSM Backend which encrypts your secrets at rest. You can read how S3 Encrypts documents with KMS [here](https://docs.aws.amazon.com/kms/latest/developerguide/services-s3.html).
+
+The highlights of SSE-KMS are:
+- You can choose to create and manage encryption keys yourself, or you can choose to use your default service key uniquely generated on a customer by service by region level.
+- The ETag in the response is not the MD5 of the object data.
+- The data keys used to encrypt your data are also encrypted and stored alongside the data they protect.
+- Auditable master keys can be created, rotated, and disabled from the AWS KMS console.
+- The security controls in AWS KMS can help you meet encryption-related compliance requirements.
+Source https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingKMSEncryption.html
+
+To configure chamber to use the S3 backend, use `chamber -b s3-kms --backend-s3-bucket=mybucket`.  You must also supply an environment variable of the KMS Key Alias to use CHAMBER_KMS_KEY_ALIAS, by default "alias/parameter_store_key" will be used.
+
+Preferably, this bucket should reject uploads that do not set the server side encryption header ([see this doc for details how](https://aws.amazon.com/blogs/security/how-to-prevent-uploads-of-unencrypted-objects-to-amazon-s3/))
+
+
+When changing secrets between KMS Keys, you must first delete the Chamber secret with the existing KMS Key, then write it again with new KMS Key.
+
+If services contain multiple KMS Keys, `chamber list` and `chamber exec` will only show Chamber secrets encrypted with KMS Keys you have access to.
+
+This feature is experimental, and not currently meant for production work.
+
 ## Null Backend (experimental)
 
 If it's preferred to not use any backend at all, use `chamber -b null`. Doing so will forward existing ENV variables as if Chamber is not in between.
--- a/store/s3storeKMS.go
+++ b/store/s3storeKMS.go
@@ -8,66 +8,61 @@ import (
 	"os"
 	"sort"
 	"time"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/s3"
 	"github.com/aws/aws-sdk-go/service/s3/s3iface"
 	"github.com/aws/aws-sdk-go/service/sts"
+	"github.com/pkg/errors"
 )
-
-const (
-	MaximumVersions = 100
-	// deprecated
-	BucketEnvVar = "CHAMBER_S3_BUCKET"
-
-	latestObjectName = "__latest.json"
-)
-
-// secretObject is the serialized format for storing secrets
-// as an s3 object
-type secretObject struct {
-	Service string                `json:"service"`
-	Key     string                `json:"key"`
-	Values  map[int]secretVersion `json:"values"`
-}
-
-// secretVersion holds all the metadata for a specific version
-// of a secret
-type secretVersion struct {
-	Created   time.Time `json:"created"`
-	CreatedBy string    `json:"created_by"`
-	Version   int       `json:"version"`
-	Value     string    `json:"value"`
-}
 
 // latest is used to keep a single object in s3 with all of the
 // most recent values for the given service's secrets.  Keeping this
 // in a single s3 object allows us to use a single s3 GetObject
 // for ListRaw (and thus chamber exec).
-type latest struct {
-	Latest map[string]string `json:"latest"`
+type LatestIndexFile struct {
+	Latest map[string]LatestValue `json:"latest"`
 }
 
-var _ Store = &S3Store{}
+type LatestValue struct {
+	Version   int       `json:"version"`
+	Value     string    `json:"value"`
+	KMSAlias  string    `json:"KMSAlias"`
+}
 
-type S3Store struct {
+var _ Store = &S3KMSStore{}
+
+type S3KMSStore struct {
 	svc    s3iface.S3API
 	stsSvc *sts.STS
 	bucket string
 }
 
-// Deprecated; use NewS3StoreWithBucket instead
-func NewS3Store(numRetries int) (*S3Store, error) {
+// Deprecated; use NewS3KMSStoreWithBucket instead
+func NewS3KMSStore(numRetries int) (*S3KMSStore, error) {
 	bucket, ok := os.LookupEnv(BucketEnvVar)
 	if !ok {
 		return nil, fmt.Errorf("Must set %s for s3 backend", BucketEnvVar)
 	}
 
-	return NewS3StoreWithBucket(numRetries, bucket)
+	return NewS3KMSStoreWithBucket(numRetries, bucket)
 }
 
-func NewS3StoreWithBucket(numRetries int, bucket string) (*S3Store, error) {
+func(s *S3KMSStore) KMSKey() string {
+	fromEnv, ok := os.LookupEnv("CHAMBER_KMS_KEY_ALIAS")
+	if !ok {
+		return DefaultKeyID
+	}
+	if !strings.HasPrefix(fromEnv, "alias/") {
+		return fmt.Sprintf("alias/%s", fromEnv)
+	}
+
+	return fromEnv
+}
+
+func NewS3KMSStoreWithBucket(numRetries int, bucket string) (*S3KMSStore, error) {
 	session, region, err := getSession(numRetries)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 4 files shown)


nickatsegment:
The method seems pretty sound to me.

I wonder if you'd be receptive to the idea of making this a new backend? Considering the index format needs to change. And then you could maybe put it in one file?

matthewborden:
Thanks for taking a look @nickatsegment.

> I wonder if you'd be receptive to the idea of making this a new backend? Considering the index format needs to change?

No problem. I'll move it into an "S3-KMS" backend.

> And then you could maybe put it in one file?

Just checking if you were referring to the individual secret files (storing the history in the summary "latest" file and not having the individual files) or storing only one summary file? For the latter, S3 requires each file to be encrypted with a single KMS key so I don't think it will be possible to combine them. Open to any ideas of how to work around having multiple summary files though!

nickatsegment:
> For the latter, S3 requires each file to be encrypted with a single KMS key
Yep, good point. Keep it the way it is.

I wonder how old versions would play with the encryption key. Say your current key got compromised, so you rolled the key and published a new version, including the summary file. Those are safe, but the old version is still hanging around having been encrypted with a compromised key.

matthewborden:
> I wonder how old versions would play with the encryption key. Say your current key got compromised, so you rolled the key and published a new version, including the summary file. Those are safe, but the old version is still hanging around having been encrypted with a compromised key.

Good point. The old summary file for the compromised key would still be present.

Are you able to fill me in on the context for the summary file? Was it to make `List` and `Exec` faster?
It definitely adds a layer of complexity to this. We could potentially not use the summary file and instead use ListObjects to index the chamber keys in the service and then fetch them in parallel?

nickatsegment:
Yep, just to make getting the latest secrets possible in one lookup. The alternative is listing all objects and then fetching the versions of each secret that's the latest. If you've got lots of version or lots of secrets, that could end up being a fair amount of listing.

I think there was another concern that you'd want a consistent snapshot. You can imagine that there's a potential race condition between writers and readers when there's multiple objects involved on both sides. `__latest.json` solves that, since it will always be consistent, if not up to date.

matthewborden:
That all sounds perfectly reasonable. I'll continue with my current plan of merging multiple `KMS_PREFIXED_latest.json` files together and handling when users only have one of the KMS keys.

In terms of consistency, I was looking at S3 Object Locking and thought it would be useful to ensure that only one person is able to write to the latest file at a time. https://docs.aws.amazon.com/AmazonS3/latest/dev/object-lock.html

matthewborden:
One issue I’ve run into is when your writing an existing secret using a new key. I think we should prevent this and explode with a message that asks you to delete the existing Chamber key with the old KMS key, and then write it again with the new KMS key.

matthewborden:
Hey @nickatsegment, would love your thoughts on the implementation when you get a chance!

It's my first time writing go, so feel free to point our any faux pas that I've made along the way.

README.md
```diff
@@ -219,6 +219,29 @@ To configure chamber to use the S3 backend, use `chamber -b s3 --backend-s3-buck
...
+
+The highlights of SSE-KMS are:
+- You can choose to create and manage encryption keys yourself, or you can choose to use your default service key uniquely generated on a customer by service by region level.
+- The ETag in the response is not the MD5 of the object data.
+- The data keys used to encrypt your data are also encrypted and stored alongside the data they protect.
+- Auditable master keys can be created, rotated, and disabled from the AWS KMS console.
+- The security controls in AWS KMS can help you meet encryption-related compliance requirements.
+Source https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingKMSEncryption.html
+
+To configure chamber to use the S3 backend, use `chamber -b s3-kms --backend-s3-bucket=mybucket`.  You must also supply an environment variable of the KMS Key Alias to use CHAMBER_KMS_KEY_ALIAS, by default "alias/parameter_store_key" will be used.
```
nickatsegment: How would you feel about making this a flag instead? I just think being explicit with these security details is better. I realize it doesn't mesh with the existing usage of `CHAMBER_KMS_KEY_ALIAS`, but I think as our scope grows, the "simplicity" of env vars over flags diminishes.

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
...
+
+var _ Store = &S3KMSStore{}
+
+type S3KMSStore struct {
+	svc    s3iface.S3API
+	stsSvc *sts.STS
+	bucket string
+}
+
+// Deprecated; use NewS3KMSStoreWithBucket instead
```
nickatsegment: You can just delete this whole function I think :)

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
...
+	bucket, ok := os.LookupEnv(BucketEnvVar)
+	if !ok {
+		return nil, fmt.Errorf("Must set %s for s3 backend", BucketEnvVar)
+	}
+
+	return NewS3KMSStoreWithBucket(numRetries, bucket)
+}
+
+func (s *S3KMSStore) KMSKey() string {
+	fromEnv, ok := os.LookupEnv("CHAMBER_KMS_KEY_ALIAS")
```
nickatsegment: We break this rule elsewhere, but I would really prefer to not read env vars in lib code. It makes testing messy, and IMO it breaks encapsulation.

Since this is optional, I think the solution would be a new public field on the struct. If unset, use the default. In the UI code, set the field based on the env var (or flag, as I suggested elsewhere).

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+func (s *S3KMSStore) Write(id SecretId, value string) error {
+	index, err := s.readLatest(id.Service)
+	if err != nil {
+		return err
+	}
+
+	if val, ok := index.Latest[id.Key]; ok {
+		if val.KMSAlias != s.KMSKey() {
```
nickatsegment: I wonder if this comparison is robust enough? You could have multiple aliases referring to the same key, right? Not sure what's desirable here.

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
...
+			return errors.New(fmt.Sprintf("You're attempting to write a Chamber key with a different KMS Key to the one it currently uses. Instead delete the existing chamber key using the current KMS key (%s) and then write the secret using the new KMS Key.", val.KMSAlias))
```
nickatsegment: Use `fmt.Errorf` :)

Also, this error string is a little long/prose-y. I'm pretty sure the `store` package isn't used too much outside of the chamber CLI (especially `Write`), but this sounds more like an error you'd show to a user than a log. I suggest something terse like `Unable to overwrite secret %s using new KMS key %s; mismatch with existing key %s`.

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
...
+}
+
+func (s *S3KMSStore) Delete(id SecretId) error {
+	index, err := s.readLatest(id.Service)
+	if err != nil {
+		return err
+	}
+
+	if val, ok := index.Latest[id.Key]; ok {
+		if val.KMSAlias != s.KMSKey() {
```
nickatsegment: Same comments as with `Write`

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
+package store
```
nickatsegment: A bunch of this code is duplicated from the S3 store (obviously :P). I wonder if we could cut down on the duplication by embedding an S3Store in S3KMSStore and overriding functions where there's a difference? Hopefully with 0 refactoring of S3Store

https://medium.com/random-go-tips/method-overriding-680cfd51ce40 for some tips

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
...
+		return DefaultKeyID
+	}
+	if !strings.HasPrefix(fromEnv, "alias/") {
+		return fmt.Sprintf("alias/%s", fromEnv)
+	}
+
+	return fromEnv
+}
+
+func NewS3KMSStoreWithBucket(numRetries int, bucket string) (*S3KMSStore, error) {
```
nickatsegment: The S3 store's long name is to maintain compatibility. You're starting fresh here.

```
type S3KMSStore struct {
	svc    s3iface.S3API
	stsSvc *sts.STS
	bucket string
	kmsKeyAlias string
}

func NewS3KMSStore(numRetries int, bucket string, kmsKeyAlias string) (*S3KMSStore, error) {
	if kmsKeyAlias == "" {
		kmsKeyAlias = DefaultKeyID
	}
	...
```
Then we can get rid of `KMSKey()` as well.

store/s3storeKMS.go
```diff
@@ -0,0 +1,486 @@
...
+
+	if err != nil {
+		return latestResult, err
+	}
+
+	return latestResult, nil
+}
+
+func (s *S3KMSStore) latestFileKeyNameByKMSKey() string {
+	return fmt.Sprintf("__kms_%s__latest.json", strings.Replace(s.KMSKey(), "/", "_", -1))
```
nickatsegment: `KMSKey()` could be an alias (and usually is). What happens when you have multiple aliases pointing to the same key, and you use each of them? You'd be using different `latest` files depending on which alias you used.

README.md
matthewborden: I really like the Environment Variable interface, we run a series of different Docker containers of different security domains that we pass in CHAMBER_KMS_KEY_ALIAS into to dynamically write secrets with different keys. In practice, we could pass in the key alias and then invoke the command with the flag set to the environment variable.

Would you consider both interfaces? It can explode if you provide both the flag and the environment variable?

store/s3storeKMS.go
matthewborden: 👍

store/s3storeKMS.go
matthewborden: 👍

store/s3storeKMS.go
matthewborden: 👍

store/s3storeKMS.go
matthewborden: I was thinking the same thing, I just wasn't aware of how to override in `go`. Thanks for the tip!

store/s3storeKMS.go
matthewborden: Yeah, we would need to be able to resolve the alias to the KMS Key. It's totally fine to do that but it just adds more permissions required for using this backend.

Instead, we could read the old KMS `latest` file, delete the key from that file, write the key to the new alias file and then write both files using their respective keys.

store/s3storeKMS.go
matthewborden: See https://github.com/segmentio/chamber/pull/195#discussion_r266699280

matthewborden:
@nickatsegment Hey, I've left some responses, I probably won't start implementing any of the changes until we're agreed on the behavior.

store/s3storeKMS.go
matthewborden: 👍

README.md
nickatsegment: Yep, happy to have both, similar to how `--backend/CHAMBER_BACKEND` works.

store/s3storeKMS.go
nickatsegment: Ok, yeah, I was just thinking out loud. It's probably best to assume they're separate keys and not resolve aliases. This might mean some superfluous writes or extra `latest`s, but meh.

nickatsegment:
@matthewborden I think that's all the discussion I have for now, should be on the same page? Just `Re-request review` when you're ready

PR #210, opened 2019-07-10
update date format to include year

systemizer:
Currently, the date format printed to STDOUT only includes the month and day. This change adds the year to the date format.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -35,7 +35,7 @@ var (
 
 const (
 	// ShortTimeFormat is a short format for printing timestamps
-	ShortTimeFormat = "01-02 15:04:05"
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -35,7 +35,7 @@ var (
 
 const (
 	// ShortTimeFormat is a short format for printing timestamps
-	ShortTimeFormat = "01-02 15:04:05"
+	ShortTimeFormat = "2006-01-02 15:04:05"
```
nickatsegment: There's a small chance this might break any scripts that rely on this format, but meh.

systemizer: good point. I think this is worthwhile. we don't explicitly support an interface via stdout so whoever used it as an interface to their scripts will just have to update

PR #213, opened 2019-07-22
pipeline: use vendored go modules and ensure we always build from them

nickatsegment:


Fauzyy:
Looks like the removed of the `vendor.json` is breaking the Snyk scan. Does Snyk support go modules yet? Last time I checked I don't think it did...

nickatsegment:
> Looks like the removed of the `vendor.json` is breaking the Snyk scan. Does Snyk support go modules yet? Last time I checked I don't think it did...

It doesn't. I think there's a manual step with the sec team to get this project refreshed to ignore the vendor.json missing problem.

Issue #221, opened 2019-08-28
Feature Request: `find` command


AaronME:
In our env we have inherited a large number of services with keys duplicated throughout. Naming conventions are inconsistent. We found it useful to add a `find` command in chamber which helps us search across services for matching keys. We also implemented a flag which allows us to find keys by value, helping us locate dependencies on common resources across many services.

We believe this investigative feature would be useful to others.

```
$ chamber find -h
Find the given secret across all services

Usage:
  chamber find <secret> [flags]

Flags:
  -v, --by-value   Find parameters by value
  -h, --help       help for find
```
```
$ chamber find s3_bucket_name
Service
service1
service2
service5
```
```
$ chamber find s3://this_bucket --by-value
Service				Key
service1			s3_bucket_name
service2			s3_bucket_name
service3			s3_bucket
service4			s3_bucket_base
service5			s3_bucket_name
```

nickatsegment:
> Would be nice to support all the various backends.

I just realized your example wasn't using `s3://this_bucket` as a backend as I had assumed. Rather as a value. My bad.

PR #223, opened 2019-09-04
feat: Add `find` command

AaronME:
In response to https://github.com/segmentio/chamber/issues/221

This adds a `find` command to chamber which can be used to find similar Keys in a store.

```
$ chamber find -h
Find the given secret across all services

Usage:
  chamber find <secret> [flags]

Flags:
  -v, --by-value   Find parameters by value
  -h, --help       help for find
```
```
$ chamber find s3_bucket_name
Service
service1
service2
service5
```
```
$ chamber find s3://this_bucket --by-value
Service				Key
service1			s3_bucket_name
service2			s3_bucket_name
service3			s3_bucket
service4			s3_bucket_base
service5			s3_bucket_name
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,6 +189,19 @@ $ chamber delete service key
 including the secret's additional metadata. There is no way to recover a
 secret once it has been deleted so care should be taken with this command.
 
+### Finding
+```bash
+$ chamber find key
+```
+
+`find` provides the ability to locate which services use the same key names.
+
+```bash
+$ chamber find value --by-value
+```
+Passing `--by-value` or `-v` will search the values of all secrets and return
+the services and keys which match.
+
 ### AWS Region
 
 Chamber uses [AWS SDK for Go](https://github.com/aws/aws-sdk-go). To use a
--- a/cmd/find.go
+++ b/cmd/find.go
@@ -1,0 +1,121 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/pkg/errors"
+	"github.com/segmentio/chamber/store"
+	"github.com/spf13/cobra"
+)
+
+// findCmd represents the find command
+var findCmd = &cobra.Command{
+	Use:   "find <secret name>",
+	Short: "Find the given secret across all services",
+	Args:  cobra.ExactArgs(1),
+	RunE:  find,
+}
+
+var (
+	blankService   string
+	byValue        bool
+	includeSecrets bool
+	matches        []store.SecretId
+)
+
+func init() {
+	findCmd.Flags().BoolVarP(&byValue, "by-value", "v", false, "Find parameters by value")
+	RootCmd.AddCommand(findCmd)
+}
+
+func find(cmd *cobra.Command, args []string) error {
+	findSecret := args[0]
+
+	if byValue {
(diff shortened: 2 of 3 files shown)


pecigonzalo:
Isnt this an "covered" already doing `chamber list-services -e | grep this` ?
IMO there is a big overlap and `list-service -e` providing all output, gives the ability to filter with any other tool, and would rather leave the `grep` do filtering and do a be a bit more "UNIX" philosophy about this.

PS: Dont get me wrong, the functionality is indeed useful but I would suggest not adding many features that are easily done with other tools

AaronME:
> PS: Dont get me wrong, the functionality is indeed useful but I would suggest not adding many features that are easily done with other tools

That solution works once we know the service. However, if we are trying to figure out which services/environments are using a particular token then we would have to wrap chamber into a for loop and grep. That's certainly doable, but including the ability as a core command is less cumbersome.

pecigonzalo:
Could you provide an example? because the command I provided allows you to search for a secret without knowing its full path. eg. `chamber list-services -e | grep s3_bucket` will output the same as whats in the example in this PR.

Keep in mind `list-services` can be called with no parameters (no filter)

AaronME:
I'm not seeing `-e`  as an available flag for `list-services`.

```$ chamber list-services -e
Error: unknown shorthand flag: 'e' in -e```

```
$ chamber list-services --help
List services

Usage:
  chamber list-services <service> [flags]

Flags:
  -h, --help      help for list-services
  -s, --secrets   Include secret names in the list

Global Flags:
  -b, --backend string             Backend to use; AKA $CHAMBER_SECRET_BACKEND
                                   	null: no-op
                                   	ssm: SSM Parameter Store
                                   	s3: S3; requires --backend-s3-bucket
                                   	s3-kms: S3 using AWS-KMS encryption; requires --backend-s3-bucket and --kms-key-alias set (if you want to write or delete keys). (default "ssm")
      --backend-s3-bucket string   bucket for S3 backend; AKA $CHAMBER_S3_BUCKET
      --kms-key-alias string       KMS Key Alias for writing and deleting secrets; AKA $CHAMBER_KMS_KEY_ALIAS. This option is currently only supported for the S3-KMS backend. (default "alias/parameter_store_key")
  -r, --retries int                For SSM, the number of retries we'll make before giving up (default 10)
      --verbose                    Print more information to STDOUT
```

**Comment by pecigonzalo** [comment]
My bad, I meant `-s`

**Comment by AaronME** [comment]
Thank you! That output makes it possible to find keys with similar names across services with `grep`. That was one of the two features we wanted.

We also wanted to find identical values which might not be following naming conventions or, because of complex conventions, might have found their way into differently named variables. This is possible using bash around chamber, of course, but I prefer
```
chamber find <VALUE> -v
```
to
```
for service in `chamber list-services | grep -v Service`; do
  chamber list ${service} -e | grep <VALUE> | awk '{print ENVIRON["sevice"]"/"$1}'
done
```

pecigonzalo:
I did not notice you were also searching by value, that could have been a `--show-values` or similar parameter to `list-services` but I think `find` does a better job at not just printing all to console and only printing if it finds.
From the initial note, I thought it was searching for key/service names. 

Thanks for the explanation, certainly helpful.

PR #227, opened 2019-09-20
feat: Add `exec --strict`

nickatsegment:
Strict mode ensures that the values getting injected are the values you expect. It is off by default.

In strict mode, `chamber exec` ensures that every injected secret has a corresponding env var in the parent environment with a sentinel value (default `chamberme`).

One use case is that you if you're using chamber in a Dockerfile, you could do

```
...
ENV DB_USERNAME=chamberme
ENV DB_PASSWORD=chamberme

ENTRYPOINT ["/chamber", "exec", "--strict", "--", "/mybin"]
```

and it will crash if chamber can't find `db-username` and `db-password` secrets.

Edge cases:
- if there is a secret without an env var, the secret is not injected
- if there is an env var with a non-chamber-compatible name like `DB_password=chamberme`, this is an error, regardless of whether there is a corresponding secret

I've tested the examples and they work as stated.

I implemented support for `--strict --pristine`, but I'm not sure of its utility.

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -14,12 +14,13 @@ import (
 // When true, only use variables retrieved from the backend, do not inherit existing environment variables
 var pristine bool
 
-// When true,
+// When true, enable strict mode, which checks that all secrets replace env vars with a special sentinel value
 var strict bool
 
-// The value to expect in strict mode
+// Value to expect in strict mode
 var strictValue string
 
+// Default value to expect in strict mode
 const strictValueDefault = "chamberme"
 
 // execCmd represents the exec command
@@ -98,7 +99,6 @@ func execRun(cmd *cobra.Command, args []string) error {
 	_, noPaths := os.LookupEnv("CHAMBER_NO_PATHS")
 
 	var env environ.Environ
-	// TODO: combine these into a single LoadAll or something
 	if strict {
 		var err error
 		env = environ.Environ(os.Environ())
@@ -117,6 +117,7 @@ func execRun(cmd *cobra.Command, args []string) error {
 		for _, service := range services {
 			collisions := make([]string, 0)
 			var err error
+			// TODO: these interfaces should look the same as Strict*, so move pristine in there
 			if noPaths {
 				err = env.LoadNoPaths(secretStore, service, &collisions)
 			} else {
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/segmentio/chamber/environ"
-	"github.com/segmentio/chamber/store"
 	"github.com/spf13/cobra"
 	analytics "gopkg.in/segmentio/analytics-go.v3"
 )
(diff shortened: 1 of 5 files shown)


nickatsegment:
Looks like I need go1.13 first, will do.

nickatsegment:
https://github.com/segmentio/chamber/pull/228

cmd/exec.go
```diff
@@ -32,10 +41,32 @@ var execCmd = &cobra.Command{
...
+
+--strict will fail with unfilled env vars
+
+	$ HOME=/tmp DB_USERNAME=chamberme DB_PASSWORD=chamberme EXTRA=chamberme chamber exec --strict service exec -- env
+	chamber: extra unfilled env var EXTRA
+	exit 1
+
+--pristine takes effect after checking for --strict values
+
+	$ HOME=/tmp DB_USERNAME=chamberme DB_PASSWORD=chamberme chamber exec --strict --pristine service exec -- env
```
Fauzyy: Is it worth logging here that we're using pristine? In case someone is debugging and doesn't realize this is enabled 😄

nickatsegment: Done in https://github.com/segmentio/chamber/pull/227/commits/e6bf6b50580c81f186d8131469ea9bbbd0b1d49b. But only for verbose mode. Considering we're execing stuff here, users probably generally don't want us cluttering up the real app's stderr.

environ/environ.go
```diff
@@ -36,6 +37,31 @@ func (e *Environ) Set(key, val string) {
 	*e = append(*e, key+"="+val)
 }
 
+// Map squashes the list-like environ, taking the latter value when there are
+// collisions, like a shell would. Invalid items (e.g., missing `=`) are dropped
+func (e *Environ) Map() map[string]string {
```
Fauzyy: Bonus for a short unit test for this one, even if it's straight forward

environ/environ.go
```diff
@@ -36,6 +37,31 @@ func (e *Environ) Set(key, val string) {
...
+			// drop invalid kv pairs
+			// I guess this could happen in theory
+			continue
+		}
+		ret[s[0]] = s[1]
+	}
+	return ret
+}
+
+func fromMap(m map[string]string) Environ {
```
Fauzyy: Ditto here

environ/environ.go
```diff
@@ -82,3 +116,107 @@ func (e *Environ) Load(s store.Store, service string, collisions *[]string) erro
...
+	}
+
+	envVarKeysAdded := map[string]struct{}{}
+	for _, rawSecret := range rawSecrets {
+		envVarKey := secretKeyToEnvVarName(rawSecret.Key, noPaths)
+
+		parentVal, parentOk := parentMap[envVarKey]
+		// skip injecting secrets that are not present in the parent
+		if !parentOk {
+			continue
```
Fauzyy: Could we log this with verbose as well?

nickatsegment: Maybe, but that would mean getting `verbose` into this lib code somehow, or a logger. Not sure that's worth doing?

Fauzyy: Ah yeah that doesn't sound worth it 👍

nickatsegment: From `exec`'s point of view, I'm pretty sure this would not be surprising. If the user's got a secret in the parent that we're dropping, either:

1. User didn't want it and everything is as it should be.
2. User did want it but forgot to write a `chamberme`. Hopefully the app crashes when it can't find its DB password.

If #2 happens undetected, well, that's 2 checks chamber did, but at least it didn't inject tokens unasked for.

Issue #219, opened 2019-08-20
'chamber read' cannot find secret


milpog:
I was trying to use `chamber read` command but I have some issues. I can `read` following secrets
```
chamber read -q service test-test
chamber read -q service test_test
chamber read -q service TEST-TEST
chamber read -q service TEST_TEST
```
But following are not working
```
chamber read -q service TEST_TEST_TEST
chamber read -q service TEST-TEST-TEST
chamber read -q service DB_POSTGRES_PASSWORD
```
and returning following error message
```
Error: Failed to read: secret not found
```

milpog:
I figured it out. `chamber` is converting `key` provided by to lowercase in this line https://github.com/segmentio/chamber/blob/9d49cf8/cmd/read.go#L40. And `TEST_TEST` or `TEST-TEST` were only working because I had `test_test` and `test-test` key created earlier.

PR #238, opened 2019-11-14
Read command -- don't lowercase keys

maschwenk:
#### For the `read` subcommand

`service` makes sense to lowercase
`key` does not

Forcing the `key` to lowercase means it's impossible to read an all-caps key. This does not makes sense to me. Most ENV vars are uppercase.

#### Evidence that this is a unhelpful default:

#219
#160

### DIFF
--- a/cmd/read.go
+++ b/cmd/read.go
@@ -37,7 +37,7 @@ func read(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "Failed to validate service")
 	}
 
-	key := strings.ToLower(args[1])
+	key := args[1]
 	if err := validateKey(key); err != nil {
 		return errors.Wrap(err, "Failed to validate key")
 	}


maschwenk:
I also noticed there's no tests for the `read` command at all? I can add those if we're ok with this PR.

mdere-unbound:
Should this be applied to list, delete and write?

maschwenk:
@mdere-unbound yeah probably, hesitant to write any new code considering this hasn't been ACK'd sinc Nov.

nickatsegment:
Hi, sorry for the delay.

As I discussed in previous tickets, I have two objections:

1. This changes behaviour. Honestly I'm not sure how much impact this would have, or if people would care, but it doesn't feel safe to me. The proper fix would be to have a flag `--normalize-case` or something. Totally doable, as discussed in https://github.com/segmentio/chamber/pull/162#issuecomment-460411974. That seems like a tough and ugly task to do without breaking API compatibility somewhere, hence me closing it, but that was just a feeling, so happy to be shown I'm wrong.

2. The (current) design of chamber is that it owns and has authority over its backends -- nobody else is writing to it. Besides [this bug](https://github.com/segmentio/chamber/issues/160), chamber should be normalizing to lowercase-with-dashes on write, and so lowercasing on read should be a no-op. Obviously there is apparently a use case that runs against this ownership/normalization idea, but changing that requires some re-design and re-testing assumptions.

I've been thinking about addressing this in new version of chamber, but haven't had time. https://github.com/segmentio/chamber/wiki/Chamber3#better-guarantees https://github.com/segmentio/chamber/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+milestone%3Av3.0

maschwenk:
@nickatsegment makes sense, I'll close then

nickatsegment:
Thanks, sorry to rain on your parade, but I gotta think of everyone.

ento:
@nickatsegment Is it still the plan to support case sensitivity in v3?

We've been waiting for this PR to get merged in order to just use `read` to get the value of a specific uppercase key, instead of running `chamber exec <service> -- env` and parsing the output. Is there a recommended workaround for reading the value of an uppercase key without potentially making multiple API calls?

nickatsegment:
@ento, well I'm leaning yes, but need to evaluate. Any reason you can't just use the AWS cli? 

Also, I don't have any concrete timeline for v3.

Another alternative is using `export` and extracting with `jq` or something. I don't believe `export` changes the case.

ento:
@nickatsegment Using the AWS cli works for us.

`export` is not much different than `exec -- env` in that it'll query all parameters when we only want to get a single parameter (although `export` is much better in terms of leanness of the code involved than `exec --env`!).

The good things about being able to use `read` in a case-sensitive way are:

- Consistency of tools being used: if both chamber and AWS cli are used to interface with the parameter store, the reason why certain cases call for one of them and not the other is something to be documented and it adds a small overhead to understanding the code base.
- Less surprise: I'd expect to be able to read a value under the exact path that it's stored. (This can be mitigated with documentation.)

The way we're using chamber is that we provision SSM parameters with Terraform and bulk-edit them through a custom script. chamber is used for `exec`ing our binaries and reading specific parameter values during deploys.

Case-sensitivity will be a quality-of-life improvement and a nice-to-have, but something we'd be content with not having a known timeline.

PR #242, opened 2019-12-16
Label support for SSMStore

rh-kpatel4:
Added support for SSMStore. Format: `/foo:v1`, `/foo` service, `v1` is the label. This is backwards compatible and does not break existing API.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -107,4 +107,35 @@ func TestValidations(t *testing.T) {
 		})
 	}
 	os.Unsetenv("CHAMBER_NO_PATHS")
+
+	// Test Service format with PATH and Label
+	validServicePathFormatWithLabel := []string{
+		"foo/bar:-current-",
+		"foo.bar/foo:current",
+		"foo-bar/foo:current",
+		"foo-bar/foo-bar:current",
+		"foo/bar/foo:current",
+		"foo/bar/foo-bar:current",
+		"foo/bar/foo-bar",
+	}
+
+	for _, k := range validServicePathFormatWithLabel {
+		t.Run("Service with PATH validation and label should return Nil", func(t *testing.T) {
+			result := validateServiceWithLabel(k)
+			assert.Nil(t, result)
+		})
+	}
+
+	invalidServicePathFormatWithLabel := []string{
+		"foo:current$",
+		"foo.:",
+		"foo.bar:cur|rent",
+	}
+
+	for _, k := range invalidServicePathFormatWithLabel {
+		t.Run("Service with PATH validation and label should return Error", func(t *testing.T) {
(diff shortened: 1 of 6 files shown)


cmd/root_test.go
```diff
@@ -107,4 +107,35 @@ func TestValidations(t *testing.T) {
...
+	}
+
+	invalidServicePathFormatWithLabel := []string{
+		"foo:current$",
+		"foo.:",
+		"foo.bar:cur|rent",
+	}
+
+	for _, k := range invalidServicePathFormatWithLabel {
+		t.Run("Service with PATH validation and label should return Error", func(t *testing.T) {
```
eculver: It would be nice to add the `k` value here to the test scenario name to be able to distinguish which one failed.

store/ssmstore.go
```diff
@@ -559,3 +574,19 @@ func getChangeType(version int) ChangeEventType {
...
+
+func parseServiceLabel(serviceAndLabel string) (string, string) {
+	if labelMatchRegex.MatchString(serviceAndLabel) {
+		i := strings.Index(serviceAndLabel, ":")
+
+		if i > -1 {
+			service := serviceAndLabel[:i]
+			label := serviceAndLabel[i+1:]
+			return service, label
+		} else {
```
eculver: nit: you don't need these `else` blocks since the `if` will always return

eculver:
Can you describe the use-case and how you tested this?

cmd/root_test.go
rh-kpatel4: The error contains the failed service name in there: https://github.com/segmentio/chamber/pull/242/files#diff-ff7686b39bf90dc2520886fb874371a4R119

store/ssmstore.go
rh-kpatel4: i removed them and updated the PR

rh-kpatel4:
> Can you describe the use-case and how you tested this?

This blog from AWS describers the use of Labels to compose secrets to a label.

https://aws.amazon.com/blogs/mt/use-parameter-labels-for-easy-configuration-update-across-environments/

rh-kpatel4:
Can I get feedback on this please? Thank in advance for your time and help.

rh-kpatel4:
Just checking into see if I can get feedback on this please?

rh-kpatel4:
@eculver Could I get feedback on this please. 🙏

DanielRedOak:
Would love to see some traction on this. AWS announced yesterday support for something similar to version selection within their ECS platform using Secrets Manager, but we'd love to continue using chamber!

rh-kpatel4:
Yes, please I would also like to get some traction on this. I'm willing to make any changes requested!

PR #246, opened 2020-02-27
fix: bad regex regression introduced by #242

### LINKED ISSUE #245: Failed to validate service even though service name is valid
This appears to be a regression from #242.

Since yesterday, new builds of one of our apps started failing consistently, with an error message like:
>Error: Failed to validate service: Failed to validate service name 'my-app-name'. Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names, and colon followed by a label name

Our build jobs install chamber using the `go get` approach described in https://github.com/segmentio/chamber/wiki/Installation

We've not been using labels for our services and have no intention of using them at this time.

sworisbreathing:
Fixes #245 by making the path/label optional in the regex.
Added test coverage of the specific scenario reported in #242


sworisbreathing:
cheers. there are other issues with the regexes added in #242 but I wanted to limit the scope of my fix to just the issue I reported in #245

PR #252, opened 2020-04-09
Add debian/buster to packagecloud.io releases

mbarrien:
Debian 10.0 (codename "buster") is the most recent stable release, and was released in July 2019. This PR adds a debian buster package to those hosted on packagecloud.io, which adds compatibility for those installing chamber via apt-get on buster-based virtual machines and images.

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -14,6 +14,7 @@ endif
 
 PACKAGECLOUD_DEB_DISTROS := \
 	debian/stretch \
+	debian/buster \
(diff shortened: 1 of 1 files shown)


mbarrien:
Possible to force a release for buster to be created for the current version, or does that require a full release/version bump?

nickatsegment:
Hm, not really. Our CI scripts work off tags, and adding `buster` is a new commit and would need a new tag. I can do that, nbd

PR #265, opened 2020-05-24
feat: Add YAML support for import/export

bdwyertech:
We like YAML as you can add in-line documentation, maybe what a secret or config parameter is for.  This is especially important when you have a lot of parameters or parameters with similar names.

Being that YAML is a superset of JSON, valid JSON is valid YML so its more or less a drop-in replacement for the import function.


cmd/import.go
```diff
@@ -47,7 +47,7 @@ func importRun(cmd *cobra.Command, args []string) error {
 
 	var toBeImported map[string]string
 
-	decoder := json.NewDecoder(in)
+	decoder := yaml.NewDecoder(in)
```
nickatsegment, line 50: I'm paranoid that this switch might not produce *precisely* the same result, but I'm struggling to come up with an actual example.

nickatsegment:
Thanks @bdwyertech !

Issue #84, opened 2018-04-05
AWS Secrets Manager


mtibben:
AWS has just announced Secrets Manager
- blog: https://aws.amazon.com/blogs/aws/aws-secrets-manager-store-distribute-and-rotate-credentials-securely/
- docs: https://docs.aws.amazon.com/secretsmanager/latest/apireference/Welcome.html

This doesn't completely replace chamber though - chamber has the advantage of being a tiny binary that can sit in your docker container.

Do you think Chamber should switch backends from Parameter Store to use the Secrets Manager API directly? Or do you think chamber should stick to Parameter Store, and perhaps a new chamber-equivalent tool should fill this space?

djrodgerspryor:
As far as I can tell, secret manager is just a much expensive ($0.4 per secret/month + some per-query charges) version of parameter-store (which is free).

The only new feature is automatic rotation, which works automatically for RDS DB credentials, but it's literally just a cronjob for all other kinds of secrets (ie. you need to write your own lambda, just like you do with parameter store).

There's no reason not to support it as an alternate store backend, but _replacing_ parameter-store seems like a bad idea.

Although I'm sure it's built on-top of parameter-store internally, I just checked and you can't see secrets-manager secrets in parameter-store, so chamber would need to merge the secrets from the two APIs manually at read-time, or force the user to pick a service and only read from that.

FernandoMiguel:
When I read the blog post I thought the same, and don't understand why it's so much more expensive than store parameter

emilva:
While Parameter Store is a free service, they still charge you for KMS keys and other underlying services like CloudWatch. In fact, Secrets Manager might be cheaper than parameter store, depending on how you manage your parameters and keys.

djrodgerspryor:
You still need to pay for KMS keys to use Secrets Manager; each secret is associated with a KMS key, just like with ParameterStore SecureStrings.

jeznag:
This article argues that rate limiting is an issue with SSM: https://www.stackery.io/blog/serverless-secrets/

nickatsegment:
@jeznag 	That's 100% true and we hit it often (see https://github.com/segmentio/chamber/pull/79). It's the main motivation for the experimental s3 backend

emilva:
AWS Secrets Manager have 700 request/sec for `ReadSecret` https://docs.aws.amazon.com/secretsmanager/latest/userguide/reference_limits.html#reference_limits_rates

nickatsegment:
[...]

PR #267, opened 2020-06-13
feat: Add AWS Secrets Manager Store

danmactough:
Implements #84 (currently closed as stale)

It sounds from #84 like there's still some interest in adding support for AWS Secrets Manager as a store, but previous efforts stalled. I started working on this implementation as a proof-of-concept for work -- at a previous job, we used Chamber and liked it a lot, but at my current job, we use Secrets Manager for secrets.

I'm not entirely sure about the design of this implementation, so I'm very open to feedback. Specifically:

- All secrets for a "service" are stored in a single Secrets Manager secret. This  reduces the number of AWS API calls for the most common use cases -- `chamber exec` and `chamber env` can be fulfilled with a single API call. This should eliminate the biggest pain point I've experienced with using Chamber -- getting rate-limited during a deployment. However, there's a greater risk of hitting the 4 KiB size limit per Secrets Manager secret.
- I'm using STS to identify the user to keep the metadata on par with that available with the SSM store. Can we safely assume that STS is available in all use cases (i.e., whenever writing a secret)? Also, I'm using the user's ARN, which includes their "AssumedRole", but I'm not sure if that's equivalent to the metadata available with the SSM store.
- I haven't implemented `ListServices`. I wasn't sure what to do about the fact that we have lots of Secrets Manager secrets already, none of which are "services" that I would want to list here. Should we suggest that the service start with a prefix, such as `service/`, and limit the `ListServices` results to secrets starting with that prefix?

I feel like I should also note that I don't write Go day-to-day, so please don't hesitate to point out things that are not conventional, poor implementation, etc. 😊


nickatsegment:
> All secrets for a "service" are stored in a single Secrets Manager secret. This reduces the number of AWS API calls for the most common use cases -- chamber exec and chamber env can be fulfilled with a single API call. This should eliminate the biggest pain point I've experienced with using Chamber -- getting rate-limited during a deployment. However, there's a greater risk of hitting the 4 KiB size limit per Secrets Manager secret.

This is sort of similar to the S3 backend's index file. It's just meant as a speed hack. BTW, the S3 backend was implemented for exactly this purpose -- to avoid rate limits.

I think this is a good approach, but it limits their compatibility with non-chamber tools, which has been complained about previously WRT case. Our stance has always been that we don't shoot for non-chamber compatibility, so it's fine with me.

> I'm using STS to identify the user to keep the metadata on par with that available with the SSM store. Can we safely assume that STS is available in all use cases (i.e., whenever writing a secret)? Also, I'm using the user's ARN, which includes their "AssumedRole", but I'm not sure if that's equivalent to the metadata available with the SSM store.

Seems fine to me. I can't imagine a where someone would block `GetCallerIdentity` but allow writing to Secrets Manager.

> I haven't implemented ListServices. I wasn't sure what to do about the fact that we have lots of Secrets Manager secrets already, none of which are "services" that I would want to list here. Should we suggest that the service start with a prefix, such as service/, and limit the ListServices results to secrets starting with that prefix?

I'm fine leaving it stubbed out for now. Personally I find limited utility in `ListServices`. A prefix might be nice, but it's your call. I'd say it should be `chamber/`. We're committing to a sort of storage format version here, and much like the v1/v2 paths/no paths migration, it can be bumpy switching, so pick a good choice :)

Anyway, this all looks pretty good to me. Thanks!

nickatsegment:
I'm going to just get a second opinion from one of the others here, and then we're good to merge.

store/secretsmanagerstore.go
```diff
@@ -0,0 +1,411 @@
...
+	version := 1
+	// first read to get the current version
+	latest, err := s.readLatest(id.Service)
+	mustCreate := false
+
+	if err != nil {
+		if len(value) == 0 {
+			return err
+		}
+		if err != ErrSecretNotFound {
```
eculver, line 75: when `err == ErrSecretNotFound`, shouldn't `mustCreate` be set to `true`?

store/secretsmanagerstore.go
```diff
@@ -0,0 +1,411 @@
...
+		if err != nil {
+			return err
+		}
+	} else {
+		putSecretValueInput := &secretsmanager.PutSecretValueInput{
+			SecretId:      aws.String(id.Service),
+			SecretString:  aws.String(string(contents)),
+			VersionStages: []*string{aws.String("AWSCURRENT"), aws.String("CHAMBER" + string(version))},
+		}
+		_, err = s.svc.PutSecretValue(putSecretValueInput)
```
eculver, line 135: I'm curious if we should prevent updates to secrets that have `RotationEnabled=true` to prevent conflicts with secrets being managed by external Lambdas? Also, with regards to versions and stages, what's the rationale of introducing a `CHAMBER+version` stage? I'm wondering if it would make sense to just leave the Chamber label out to prevent complications in the future should Chamber ever want to utilize this for a rotation strategy of its own.

eculver: Also, from the docs on `PutSecretValue`:
> If another version of this secret already exists, then this operation does not automatically move any staging labels other than those that you explicitly specify in the `VersionStages` parameter

I'm wondering if we should just omit `VersionStages` and let AWS set these automatically rather than introduce potential conflicts or errors?

store/secretsmanagerstore.go
danmactough: Ah, no. This is a very tricky bit of code.

`ErrSecretNotFound` is the Chamber internal error, but `mustCreate` means that "we must create the underlying AWS Secrets Manager secret." Because this implementation doesn't have a 1:1 mapping of Chamber secret to AWS Secrets Manager secret, `ErrSecretNotFound` doesn't mean we need to create the underlying Secrets Manager secret. In fact, we will not encounter `ErrSecretNotFound` in this code path if the underlying AWS Secrets Manager secret needs to be created. If the AWS Secrets Manager secret does not exist, when we try to get it, we'll get the AWS error `secretsmanager.ErrCodeResourceNotFoundException` instead. That is the only case in which the underlying AWS Secrets Manager secret needs to be created.

Does this make sense? I'll add a code comment to explain this here.

store/secretsmanagerstore.go
danmactough: > I'm curious if we should prevent updates to secrets that have RotationEnabled=true to prevent conflicts with secrets being managed by external Lambdas? 

@eculver I'm not sure any design that supports Chamber functionality could reasonably interop with the AWS-provided Lambdas for rotating secrets. To use Secrets Manager as a Chamber store, we need to manually manage the secret's metadata together with the secret's name and value. None of the AWS-provided Lambdas for rotating secrets are going to be aware of that metadata management (and honestly, just the structure of the secret value is probably a barrier to using those AWS-provided Lambdas for rotation). Instead, if a user wishes to use the Secrets Manager store and enable automatic rotation, they'll need to use Chamber in their rotation Lambda -- meaning we wouldn't be able to prevent updates to secrets that have `RotationEnabled=true` or those rotation Lambdas won't work. 🌀

danmactough: > what's the rationale of introducing a CHAMBER+version stage

This allows us to keep more than 2 versions of a secret. Without an additional staging label, all but the `AWSCURRENT` and `AWSPREVIOUS` versions of the secret are considered "deprecated" by Secrets Manager and subject to deletion. In my experience, these get deleted pretty darn quickly.

danmactough:
> A prefix might be nice, but it's your call. I'd say it should be `chamber/`

@nickatsegment Agree that `chamber/` would be a better prefix. I'm not inclined to flesh out `ListServices` at this time if you're ok with shipping without it.

nickatsegment:
@danmactough WFM.

Just waiting on @eculver to confirm your rationale above.

danmactough:
👋 @nickatsegment @eculver Any news on this?

store/secretsmanagerstore.go
eculver: > I'm not sure any design that supports Chamber functionality could reasonably interop with the AWS-provided Lambdas for rotating secrets. 

This is exactly my point. Chamber can technically still operate on the secrets that are managed by Lambdas which would cause problems. I'm thinking that a simple check for if `RotationEnabled=true` to ensure that the lack of interoperability doesn't cause problems.

This is my only concern at this point. Everything else is looking great.

eculver:
> Any news on this?

Thanks for following up. I just want to make a decision on the `RotationEnabled=true` detection as mentioned in the comment above. That's all we have before we get this sucker merged.

danmactough:
Hey @eculver @nickatsegment FYI, I've been thinking about @eculver's comments, and I have a reworked implementation almost ready that I think will be more interoperable with rotation and other key features of Secrets Manager.

danmactough:
Ok, here's that reworked implementation. Basically, we've revamped how we store values and metadata.

Previously, the structure stored value and metadata for each secret nested under the secret's name:
image

Secrets Manager was not able to convert this structure to key/value pairs:
image

Also, this structure would have required any Rotation Lambda to understand our bespoke structure to update the value of a secret.

-----

The new structure is keeps all Chamber metadata under a single property while secrets are structured as key/value pairs.
image

Secrets Manager can understand this structure.
image

-----

I'm still tinkering with this to see if I can get Chamber to read a non-Chamber secret from Secrets Manager, which I think would be desirable. That would mean that a user could leverage Chamber solely for the utility of `chamber env` and `chamber exec` while using Secrets Manager's rotation feature without needing to know about Chamber's implementation details.

danmactough:
> I'm still tinkering with this to see if I can get Chamber to read a non-Chamber secret from Secrets Manager, which I think would be desirable. That would mean that a user could leverage Chamber solely for the utility of chamber env and chamber exec while using Secrets Manager's rotation feature without needing to know about Chamber's implementation details.

Ok @eculver @nickatsegment, I think this is ready for another review. 😅 

Now, in addition to the [change in structure](https://github.com/segmentio/chamber/pull/267#issuecomment-660787724), I've made the following additional changes to increase interoperability with non-Chamber secrets in Secrets Manager:

- slightly tweaked the behavior of the History command so that missing metadata is silently ignored 
- reworked how we unmarshal Secrets Manager JSON to handle non-string values, converting them where possible to strings

With these changes, you can create a new Secrets Manager secret using one of the "wizards" in the AWS console (such as "Credentials for RDS database") -- which will create a secret having a "port" property with a value that is a number instead of a string -- and Chamber can still read that secret. This means you can use `chamber exec` or `chamber env` with that secret even though one or more values are not strings, and even though the secret contains none of the metadata that Chamber needs for some of its other functionality (such as history and reading a specific version). You can also use `chamber write` to update these non-Chamber secrets.

The interoperability isn't 100% perfect, but it's pretty good! Nothing is "broken", at least by some definition of broken. 🙃

danmactough:
👋 @eculver @nickatsegment Please let me know if there's anything I can do to help land this -- my team is excited about adopting chamber with Secrets Manager integration. 🤗

store/secretsmanagerstore_test.go
```diff
@@ -0,0 +1,428 @@
...
+		}
+		`
+		obj, err := jsonToSecretValueObject(j)
+		assert.Nil(t, err)
+		assert.Equal(t, "database-1", obj["dbInstanceIdentifier"])
+		assert.Equal(t, "3306", obj["port"])
+		assert.Equal(t, "true", obj["isPhony"])
+		assert.Equal(t, "", obj["empty"])
+		assert.Equal(t, "", obj["nested"])
+		assert.Equal(t, "", obj["array"])
```
eculver, line 148: Is this expected? Nested values are ignored?

store/secretsmanagerstore.go
danmactough: Added a comment via 44c774e

store/secretsmanagerstore.go
danmactough: > This is exactly my point. Chamber can technically still operate on the secrets that are managed by Lambdas which would cause problems. I'm thinking that a simple check for if `RotationEnabled=true` to ensure that the lack of interoperability doesn't cause problems.

@eculver I think I understand, but let me make sure. You're saying that if `RotationEnabled=true` we can _read_ the secrets, but we may not **_write_** the secrets, correct? (And this would only be applicable if the secret already exists, of course.) This approach makes sense to me now that the data structure is read-interoperable.

One tricky bit is that we don't know value of `RotationEnabled` at this point in the code. 🤔

store/secretsmanagerstore.go
```diff
@@ -0,0 +1,545 @@
...
+				}
+			} else {
+				return err
+			}
+		}
+	}
+
+	if len(value) == 0 {
+		if mustCreate {
+			return err
```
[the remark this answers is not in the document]
danmactough: Removed via 744c0fc

store/secretsmanagerstore.go
danmactough: > I think I understand, but let me make sure. You're saying that if `RotationEnabled=true` we can _read_ the secrets, but we may not **_write_** the secrets, correct? (And this would only be applicable if the secret already exists, of course.) This approach makes sense to me now that the data structure is read-interoperable.

@eculver I went ahead and implemented this via 583cf39

danmactough:
I want to highlight this commit I just pushed cbd2cc2 It changes the behavior a bit, but I think for the better.

Currently, all keys get converted to lowercase on write and then converted to UPPERCASE when using the `env`, `exec` and `export` commands. I'm not sure why it was desirable to convert to lowercase instead of UPPERCASE on write, but I can imagine that in the case of SSM, it's more aesthetically pleasing.

When using Secrets Manager as a backend, however, it's very surprising (and annoying) to have the keys converted to lowercase because when using the AWS SDK to fetch the secret where Chamber is not available (such as in a NodeJS Lambda function), it leads to this split brain for the developer -- when I use this secret one way (with `chamber exec`), the keys are UPPERCASE but when I use this secret any other way, the keys are lowercase. Concretely, I was testing out the workflow of using Chamber to manager my secret and then using it in a NodeJS Lambda and did the following:

image

Then I realized that if I wanted to be able to refer to secrets on `process.env` the same way I do when I use Chamber, I'd have to do this:

image

It seems wrong for my Lambda to have to about implementation details of Chamber.

I think the change in cbd2cc2 -- in which the `secretsmanagerstore` converts the keys to UPPERCASE -- resolves this confusion. 

Happy to discuss.

nickatsegment:
Sorry @danmactough , so many different irons in different fires around here I've lost track :D

WRT to case, I think regardless of the reasoning for why some parts are uppercase, it would be weird if only one provider converted to uppercase, wouldn't you say? If anything, we should do the case transformation at a higher level?

danmactough:
> so many different irons in different fires around here I've lost track :D

No worries @nickatsegment. At least one of those fires has been all over the news, right? 🤑 🍾 🎊

On the case issue, I'm happy to revert cbd2cc2.

To restate my thinking a bit, I think that converting all the secret keys to lowercase when using Secrets Manager makes the interoperability that we've been striving for less useful. Imagine that you use Chamber to manage a secret in Secrets Manager and when you use `chamber env` or `chamber exec` the environment variable names are always UPPERCASE, but then, expecting Chamber and Secrets Manager to be interoperable, in another context (where using Chamber is not possible or practical) you use Secrets Manager to get the secret and the property names are all lowercase instead of UPPERCASE -- I think that's surprising in a bad way.

But honestly, normalizing the names in ANY way is _kind of_ surprising. Like, why can't I have PascalCase or camelCase environment variable names if that's what I want? So, like you say maybe consistency is better at this point, even if it's not optimal, and perhaps an improvement in v3 😄 would be to find a way to preserve the original input for environment variable names.

danmactough:
rebased and dropped cbd2cc2

nickatsegment (approved):

store/secretsmanagerstore.go
```diff
@@ -0,0 +1,562 @@
+package store
```
nickatsegment, line 1: Last thing, I promise :)

Would you be willing to be the maintainer of this backend for the time-being? Nothing formal or enforceable obviously. All PRs would still go through Segmenters for approval (we haven't quite figured out how to deal with giving external folks merge access). We'll just defer to you on issues and PRs.

If that sounds good, just add a comment to the top of this file:

```suggestion
// Secrets Manager Store is maintained by Dan McTough https://github.com/danmactough. Thanks Dan!
package store
```

nickatsegment:
OK, looks good to me. Just that question on notating maintainership :) Then we'll merge

store/secretsmanagerstore.go
danmactough: No problem. I'll just correct the spelling of my name. 🙃

danmactough: Added via 53be3ab @nickatsegment

nickatsegment:
Re interoperability, basically chamber is not really designed to be interoperable: it's designed to work with secrets it has written, and makes no attempt to make its write format flexible or interoperable.

Making it interoperable is perhaps a nice idea, but obviously would be a big change and require a bunch more knobs and dials that are all opportunities for error. In my mind, it's fundamentally a different tool at that point.

> But honestly, normalizing the names in ANY way is kind of surprising. Like, why can't I have PascalCase or camelCase environment variable names if that's what I want?

I believe the intention is to reduce the possibility of error and confusion. It's a tradeoff of flexibility in the name of simplicity and avoiding footguns.

store/secretsmanagerstore.go
nickatsegment: > No problem. I'll just correct the spelling of my name. 🙃

Whoops! My bad, sorry.

PR #270, opened 2020-06-28
Fix directions for setting env vars, close #259

### LINKED ISSUE #259: Docs: Correct directions for terminal env vars
The directions for setting env vars in your terminal given in #247 are wrong. 

https://github.com/segmentio/chamber/blob/c2a053008714ca9c1aede3e1a804ae5764cb96b7/README.md#L174-L178

This does not work correctly with various special characters. In general, use of the `dotenv` format should be avoided (see #208), and especially here when we have #184 specifically to handle this use case. 

Recommended change:

To set env vars in your terminal you can use the `chamber env` command. For example, 
```shell
source <(chamber env service)`
printf "%s" "$SERVICE_VAR"
```

Nuru:
Fix directions for setting env vars, close #259

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,10 +172,10 @@ file formats are supported:
 File is written to standard output by default but you may specify an output
 file.
 
-To set env vars in your terminal you can eval the dotenv export format. For example, 
+To set env vars in your terminal you can use the `chamber env` command. For example, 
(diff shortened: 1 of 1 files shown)


nickatsegment:
Thanks!

Issue #272, opened 2020-07-01
Configure aws sdk retryer options along with retry attempts


kespinola:
### Goal
Able to configure retry delay parameters along with MaxRetries on aws sdk using command line arguments.

### Today
The chamber cli allows operators to set retries but not the retry delay parameters of the aws sdk.

### Issue
My org has a large fleet of instances and performs blue-green deploy strat which currently causes us to hit advanced tier throttle limits of 1k requests/per second. 

### Desire
Our team would like to set retry delay parameters so we can retry between 1-5 sec after a failure.

https://github.com/segmentio/chamber/blob/bcfb2fd1d1a4e9bd22fb0f057bbe6073933945c9/vendor/github.com/aws/aws-sdk-go/aws/client/default_retryer.go#L16-L55

https://github.com/segmentio/chamber/blob/99008c1dab17d57a6bede49ac3f3c6d264c299f2/store/shared.go#L34-L40

### References
https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config

kespinola:
The operator is now able to set `--min-throttle-delay` when invoking chamber exec using ssm as the backend.

PR #273, opened 2020-07-01
feat: Configure aws sdk min throttle delay using command argument

kespinola:
Resolution to #272

### DIFF
--- a/store/ssmstore.go
+++ b/store/ssmstore.go
@@ -6,9 +6,11 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/client"
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/aws/aws-sdk-go/service/ssm/ssmiface"
 )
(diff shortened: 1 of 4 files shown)


store/ssmstore.go
```diff
@@ -40,15 +42,17 @@ type SSMStore struct {
...
+func NewSSMStore(numRetries int, minThrottleDelay time.Duration) (*SSMStore, error) {
 	ssmSession, region, err := getSession(numRetries)
 	if err != nil {
 		return nil, err
 	}
 
+	retryer := client.DefaultRetryer{NumMaxRetries: numRetries, MinThrottleDelay: minThrottleDelay}
+
 	svc := ssm.New(ssmSession, &aws.Config{
-		MaxRetries: aws.Int(numRetries),
```
kespinola, line 50: Setting max retries on the retryer instead of config because config document notes max retries is ignored if retryer and it are specified.

store/ssmstore.go
```diff
@@ -40,15 +42,17 @@ type SSMStore struct {
 }
 
 // NewSSMStore creates a new SSMStore
-func NewSSMStore(numRetries int) (*SSMStore, error) {
+func NewSSMStore(numRetries int, minThrottleDelay time.Duration) (*SSMStore, error) {
```
nickatsegment, line 45: This is part of the public API and cannot be changed without a version bump. Please create a new function `NewSSMStoreWithMinThrottleDelay`

store/ssmstore.go
nickatsegment: Link to doc?

kespinola: https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config

Thank you for prompting the second look. In rereading it the top-level MaxRetries option takes precedence. Just setting on the retryer does preserve behavior but my original comment is incorrect.

kespinola:
Thank you for the initial review @nickatsegment. I've resolved your comments.

nickatsegment:
Great, I'm gonna merge. Any chance you've done any (manual) tests to show the new behaviour working?

kespinola:
I haven't but I will I'll test it tomorrow on the app which original prompted the change.

PR #287, opened 2020-11-19
WIP: Adding include feature to chamber

jdwilly2001:
**Use Case**
We have several services which have to use some secrets which are duplicated around.  Wanted to create an ability for a secret store to include other secret stores at runtime. This will reduce duplication of secrets and allow separate groups to manage sets of secrets via IAM roles.

**Proposed Solution**
Add new feature for the Secret Store interface for WriteInclude, RemoveInclude. Leave it up to the store on exact method of implementation, but any secret stores mentioned in the includes should be brought back by Read/List/ListRaw methods.

Implement a new type of ServiceMetadata to hold processing instructions for a particular service. 

**SSM Store Support for Includes**
Adding SSM store support for includes. Implemented by using a 'metadata' key in ParaemterStore that will be used to house any operational info about a service. The metadata is optional and only exists when the WriteInclude function is called for the service. If the metadata does not exist, the program will continue without error. 

The metadata includes a property "IncludedServices" which is a string[] of services to include. Upon reading/listing secrets for a service, a recursive call with the included service will be made to include secrets. Currently no check/limitation on included secrets.

**List command updated to reflect service**
If a particular key does not directly live under the service for which List was called, it will return the owning service name in parenthesis as an indicator that it is owned/maintained in another service.

**Open Issues/Questions**

- How should export function handle secrets?
- Guard against cyclical dependencies (Service A -> Service B -> Service A)
- Implement other functions

### DIFF
--- a/cmd/include.go
+++ b/cmd/include.go
@@ -1,0 +1,47 @@
+package cmd
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/segmentio/chamber/v2/store"
+	"github.com/spf13/cobra"
+)
+
+var (
+	includeCmd = &cobra.Command{
+		Use:   "include <service> <another-service>",
+		Short: "create a run-time include. see documentation",
+		Args:  cobra.ExactArgs(2),
+		RunE:  include,
+	}
+)
+
+func init() {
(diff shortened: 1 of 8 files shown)


jdwilly2001:
Closing this, created on accident.

PR #290, opened 2020-11-30
Fix go 1.15 build

kalekseev:
Fix as suggested by error message

```
# github.com/segmentio/chamber/v2/store
store/secretsmanagerstore.go:216:78: conversion from int to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)
FAIL    github.com/segmentio/chamber/v2/store [build failed]
```


nickatsegment:
cc @danmactough

I'm not really sure if this would have any behaviour-breaking consequences. I don't know if we use the version for anything but info.

danmactough (approved):
Thanks @kalekseev

danmactough:
> I'm not really sure if this would have any behaviour-breaking consequences. I don't know if we use the version for anything but info.

No worries here @nickatsegment

PR #292, opened 2021-02-02
cmd: use x/sys/unix.Exec

kevinburkemeter:
We're not supposed to use the syscall package directly, and unix.Exec
does a better job of ensuring that the appropriate command is issued
for each build/OS combination.

I believe it was written this way at first because the unix package
did not have an "Exec" function, which was an oversight that was
corrected at least a year ago.


emmy-byrne-segment:
I like the change but looks like the mod file didn't get updated perhaps?

```
build github.com/segmentio/chamber/v2: cannot load golang.org/x/sys/unix: open /go/src/github.com/segmentio/chamber/vendor/golang.org/x/sys/unix: no such file or directory
make: *** [Makefile:19: test] Error 1
```

kevinburkemeter:
Whoops - should be fixed now.

emmy-byrne-segment:
Did you intend to commit the vendor folder? Will this not pull it down during build?

kevinburkemeter:
> Did you intend to commit the vendor folder? Will this not pull it down during build?

I'm confused what you mean... the vendor directory is checked in, I added a dependency on x/sys/unix, so I checked it into the vendor directory. This is, as far as I can tell, how the other vendored dependencies got added, for example here. 32c22da5ab7faa932516d22654cf1059e694e4a3

eculver:
Yeah, this all looks good. I think there was just some confusion around re-vendoring things. Thanks for adding!

eculver (approved):

PR #298, opened 2021-03-20
feat: Support tagging of AWS resources

### LINKED ISSUE #89: Feature Request: Support AWS Tags
A great feature to see would be the ability to add tags to secrets with command line flags when using the `write` command. In our organization we use tags to quickly see all resource usage on a project-per-project basis (for projects that share our common account and don't warrant their own account) using AWS System Manager resource groups.

This is possible using the [`AddTagsToResource` Systems Manager API call](https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_AddTagsToResource.html).

Syntax would be something like:

```
$ chamber write my_service my_key my_value --tag tag1=value1 --tag tag2=value2
```

I'll try to learn me some go and see if I can whip up a PR when I find the time but thought I'd document this here and see if there's any interest.

### LINKED ISSUE #297: Support tagging of created resources
Hi,

I needed tags on all the SSM parameters that chamber created in my AWS account, so I added support for a --tags-file \<path\> flag on my fork.
Do you think this would be worth merging if I tidied it up and implemented it for the other stores?

Thanks,
Janos

Aresius423:
Please, let me know how I could improve the logic, the formatting, the structuring, etc.

resolves #89, resolves #297


eculver:
Sorry for delay in addressing this. I'll review this ASAP.

common/tags.go
```diff
@@ -0,0 +1,168 @@
+package common
```
eculver, line 1: This appears to be isolated to provide tagging functionality, so rather than `common`, I'm curious if we should call this  package `tagging`? The APIs therein would be more descriptive too (eg. `tagging.FromSSM()`). Thoughts?

common/tags.go
```diff
@@ -0,0 +1,168 @@
...
+
+func AddCommandWithTagging(parentCmd *cobra.Command, cmd *cobra.Command) {
+	// recommended way to set up tagging - prevents accidental override of pre-run hooks
+	SetupTagging(cmd)
+	parentCmd.AddCommand(cmd)
+}
+
+
+func SetupTagging(cmd *cobra.Command) {
+	// Set up flags required for tagging
```
eculver, line 35: looks like this needs to have `go fmt` run on it.

common/tags.go
```diff
@@ -0,0 +1,168 @@
...
+var (
+	tagsFilePath string
+	tags map[string]string
+	argTags []string
+	tagsString string
+	ssmTags []*ssm.Tag
+	secretsManagerTags []*secretsmanager.Tag
+)
+
+func AddCommandWithTagging(parentCmd *cobra.Command, cmd *cobra.Command) {
```
eculver, line 27: The wiring to the CLI should be confined to the `./cmd` package. I'd suggest we adjust some of the exported methods of this package to support this change. For example, if instead of `loadTagsFile` there was a `tagging.Config` type (or something similar) with proper `yaml` struct tags, the code in `./cmd` could use this type to unmarshal from a file when it is read from the command-line. The primary concern here is that the code here is highly coupled to a few other packages which makes testing it challenging. 

I'd suggest a refactor to make all the functionality in this package testable independently.

common/tags.go
Aresius423: Sounds good! This is such a minor feature that I didn't know if a separate package for it would be warranted.

Aresius423:
I moved tags.go to the tagging package.
I initially went with offloading the command registration and patching PreRun because I couldn't find a way to create an unnamed command that I could make the parent of both _import_ and _write_ and not break the existing command structure.

Now _EnsureTagsLoaded()_ is called explicitly from the import and write functions to populate the private tags map if it hasn't been done yet, and as before, the getter functions assemble the store-specific data structure the first time they're requested.

Aresius423:
Hi,

could we pick this up and see what still needs to be done before we can get it merged?

PR #307, opened 2021-06-04
Add support for linux/arm64

Pryz:
I need to run chamber on some AWS Graviton instances so adding the support to release a arm64 binary.

PR #311, opened 2021-08-04
Linux arm64 release fix for github and packagecloud

sindhusegment:
This is a fix for PR https://github.com/segmentio/chamber/pull/307 so that we can release linux arm64 chamber to github as well as an arm64 debian to packagecloud

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -158,9 +158,9 @@ publish-dockerhub:
 	docker push segment/chamber:$(DOCKERHUB_TAG_PREFIX)$(VERSION_MAJOR)
 	docker push segment/chamber:$(DOCKERHUB_TAG_PREFIX)$(VERSION_NO_V)
 
-dist: dist/chamber-$(VERSION)-darwin-amd64 dist/chamber-$(VERSION)-linux-amd64 dist/chamber-$(VERSION)-windows-amd64.exe dist/chamber_$(VERSION)_amd64.deb dist/chamber_$(VERSION)_amd64.rpm dist/chamber-$(VERSION).sha256sums
+dist: dist/chamber-$(VERSION)-darwin-amd64 dist/chamber-$(VERSION)-linux-amd64 dist/chamber-$(VERSION)-linux-arm64 dist/chamber-$(VERSION)-windows-amd64.exe dist/chamber_$(VERSION)_amd64.deb dist/chamber_$(VERSION)_amd64.rpm dist/chamber-$(VERSION).sha256sums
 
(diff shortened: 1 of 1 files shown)


asaf-erlich (approved):
LGTM - double check the circle job after it is merged

asaf-erlich (approved):
approved again with latest commit

Makefile.release
```diff
@@ -154,15 +169,20 @@ dist/nfpm-$(VERSION).yaml: | dist/
 dist/chamber_$(VERSION)_amd64.deb: dist/nfpm-$(VERSION).yaml dist/chamber-$(VERSION)-linux-amd64
 	nfpm -f $< pkg --target $@
 
+dist/chamber_$(VERSION)_arm64.deb: dist/nfpm-$(VERSION).yaml dist/chamber-$(VERSION)-linux-arm64
```
sindhusegment, line 172: @asaf-erlich do you think I need to add an nfpm dist for arm64 (the same as line 166 & 167 for amd64)? Not entirely sure what this is doing

asaf-erlich: The best way to test is to run the command and confirm all the release packages end up in the correct place and work as expected, but it's a chicken-and-egg problem because we probably don't want to do that when it's not yet merged into master. Probably it would be nice to add some kind of dry run flag for testing, but if it's missing I'm personally ok (without being told by anyone else) merging it and fixing it if it's broken quickly.

I'll give it a bit of a look-over but I would assume you need one for each case, especially if there's a reference to it somewhere then it would simply fail without it. I had to look up and confirm what (nfpm)[https://github.com/goreleaser/nfpm] is and it makes sense to me you need one for arm64. 

So yeah dist/chamber_$(VERSION)_arm64.deb is referenced in your PR so if you don't add this the make command will fail.

asaf-erlich: Ok I didn't understand your question, yes you probably need it.

asaf-erlich: Ok definitely, you're referring to `dist/chamber-$(VERSION)-linux-arm64` right? that is being referenced but no make command exists for it, so it'll fail.

sindhusegment: yeah there's no nfpm with linux-arm64. Would adding 
	
`dist/nfpm-$(VERSION)-arm64.yaml: | dist/
	sed -e "s/\$${VERSION}/$(VERSION)/g" -e "s|\$${DIST_BIN}|dist/chamber-$(VERSION)-linux-arm64|g" < nfpm.yaml.tmpl > $@`

and then

`dist/chamber_$(VERSION)_arm64.deb: dist/nfpm-$(VERSION)-arm64.yaml dist/chamber-$(VERSION)-linux-arm64
	nfpm -f $< pkg --target $@`

be sufficient?

sindhusegment: Yeah it's unfortunate that I it's easiest to test once merging it in 😕

asaf-erlich: I believe so, but without testing it all the way to the end I'm not 100% sure.

sindhusegment: okay, I have just pushed that change through and if you are okay, I will merge to master and see what happens?

asaf-erlich: 🚢  and just keep on eye on it and be ready to fix it

PR #312, opened 2021-08-05
Add circle building on our test branch so we can reproduce in circle

asaf-erlich:


asaf-erlich:
Going to go ahead and close this PR assuming it was only for temporary testing.

PR #313, opened 2021-08-19
Revert "Linux arm64 release fix for github and packagecloud"

sindhusegment:
Reverts segmentio/chamber#311

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -30,9 +30,9 @@ PACKAGECLOUD_RPM_DISTROS := \
 
 publish: publish-github publish-packagecloud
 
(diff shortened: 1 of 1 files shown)


asaf-erlich (approved):

masoncj:
Would you mind describing why this was reverted?  It would be great to have arm64 support!

asaf-erlich:
@masoncj It's been a while but the TLDR is that it was not successfully building chamber for arm 64 correctly and uploading it. I don't remember all the details. @emmy-byrne-segment do you remember if we ended up building chamber manually separately as a one-time thing to push things through? If we aren't building it now for arm 64 it should probably be on our roadmap to support it at some point.

PR #316, opened 2021-09-14
Fix copiable code snippets in README.md

msantl:
Using the Copy button that Github renders on the code snippets in README.md, ie.
Screenshot 2021-09-14 at 14 52 01
the content copied to the clipboard is `$ go get github.com/segmentio/chamber`.

Pasting this content won't work because it's prefixed with the `$` character.

In this PR, I'm removing all the `$` characters from the copiable code snippets so that the Copy button copies the right content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ $ CHAMBER_NO_PATHS=1 chamber export foo | chamber import foo -
 If you have a functional go environment, you can install with:
 
 ```bash
-$ go get github.com/segmentio/chamber
+go get github.com/segmentio/chamber
 ```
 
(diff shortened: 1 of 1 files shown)


asaf-erlich (approved):

msantl:
@asaf-erlich Thanks for the quick review. In the meantime, I realized that there are more code snippets that have the same issue so I went ahead and updated all of those. I'll also update the PR summary with all the changes. 

Please take another look and feel free to merge the PR (if you think it's good to go).

asaf-erlich (approved):

PR #320, opened 2021-12-07
Add an environment variable for the number of retries.

estheruary:
My use case is wanting to set a global retry policy across lots different chamber calls. Tried my best to follow the style of the project. Let me know if I need to change anything.


estheruary:
Keeping it alive!

mckern (approved):

PR #328, opened 2022-02-09
Feature/tagging

Aresius423:
Reopening #298


barrerajl:
This would help us a lot.

filiptepper:
Anything I can help to get this merged?

mckern:
👋🏻 Hi @Aresius423 -- I'm sorry that we let this sit for _so, so **long**_. We provided you with a bad experience as a first time contributor, and we're sorry we did not reply to you about this sooner.

This is an interesting feature, and I've taken a look at what it takes to get it rebased against `master` -- in that regard, the good news is that it still applies pretty cleanly (with a small handful of things needed to be addressed in merge conflicts), and I'm happy to push the rebased branch if you still want to get this merged; I didn't want to unilaterally push to your branch/fork. But, I have some concerns/feedback that you should keep in mind:

- 175 new lines of code without any tests at all is a tough maintenance burden to say yes to. Since you're introducing a new package, that should probably have accompanying tests. There's also no accompanying tests for `store/s3store.go`, `store/s3storeKMS.go`, `store/secretsmanagerstore.go`, or `store/ssmstore.go`. Of those, `secretsmanagerstore_test.go` and `ssmstore_test.go` already exist -- so at a minimum I'd like to see some test validation that _existing_ behavior is maintained along with test validation for the new behavior you're introducing.
- I am not sure how this interacts with/observes the guidelines laid out in AWS's docs about [tagging keys](https://docs.aws.amazon.com/kms/latest/developerguide/tagging-keys.html) -- this is some place where tests for the new behavior would be invaluable for assessing the maintenance cost of merging this. In the absence of tests, I'd at least like to see some code comments/README guidelines around that.
- Reading tags from `os.Stdin` in `loadTagsFile()` seems like it may end up conflicting with how `write` and `import` subcommands already read from `os.Stdin`; I get that there's scenarios where one "path" reads from a file, while another reads from `os.Stdin` but overall I think this could be simplified by exclusively allowing reading tags from a file or from command-line arguments.

I may have some other notes after reviewing this, **if** you still want to merge this. Based on its age, I think it's entirely reasonable if you'd rather close this. How would you like to proceed?

Aresius423:
Hey!

No worries.
Let's close it for now - I don't currently have the capacity to address the very valid concerns you've raised.
If I ever get to pick this up again, or someone else wants to have a go at it, we can open a new PR.

PR #330, opened 2022-02-23
fix x509: certificate signed by unknown authority error when using docker

1david5:
This fixes the below error when using chamber in docker: `docker run segment/chamber:2.10.8 list-services ...`

```
Error: Failed to list store contents: RequestError: send request failed
caused by: Post "https://ssm.us-east-1.amazonaws.com/": x509: certificate signed by unknown authority
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,9 +6,8 @@ COPY . .
 ARG VERSION
 RUN test -n "${VERSION}"
 
-RUN apk add -U make
+RUN apk add -U make ca-certificates
 RUN make linux VERSION=${VERSION}
-RUN apk --no-cache add ca-certificates
 
 FROM scratch AS run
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -8,9 +8,11 @@ RUN test -n "${VERSION}"
 
 RUN apk add -U make
 RUN make linux VERSION=${VERSION}
+RUN apk --no-cache add ca-certificates
```
pop, line 11:
```suggestion
RUN apk add -U make ca-certificates
RUN make linux VERSION=${VERSION}
```

We can do both installs in one line.

pop:
I have also encountered this issue and would love to see this fix merged.

alecjacobs5401 (approved):
Thanks for this fix!

LGTM 👍

PR #332, opened 2022-03-15
Convert env subcommand to use ListRaw

nosammai:
Switches env subcommand to use `ListRaw` and GetParametersByPath from #79


IcyWednesdays (approved):

mckern (approved):
Normally I'd ask for a supporting test/suite of tests but it looks like env.go is unfortunately test-free... quickly looking at a local example of before/after behavior and not seeing any difference in the output (which I will consider a reasonable indicator of function here).

PR #335, opened 2022-04-05
Transfer Circle CI Automation to Github Actions

sindhusegment:
### Changes

- Add github actions workflows to build, test and release chamber
- Remove circleci pipeline config and references to that in the project
- Remove package cloud from github actions pipeline

### TODO

- Add docker hub credentials to repository settings ✅ 
- Verify if segment relies on installation from package cloud ✅ - Filed a ticket  
- Test release on tag creation (once merged to master)
- Remove package cloud from installation docs (after merged and tested in actions)
- Remove circleci from repo webhook settings (after merged and tested in actions)


.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
+name: Release
+on:
+  push:
+    tags: 
+     - /v[0-9]+(\.[0-9]+)*(-[a-zA-Z0-9-]+)?/
```
rikez, line 5:
```suggestion
     - 'v[0-9]+(\.[0-9]+)*(-[a-zA-Z0-9-]+)?'
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+    - name: Login to DockerHub
+      uses: docker/login-action@v1
+      with:
+        username: ${{ secrets.DOCKER_USER }}
+        password: ${{ secrets.DOCKER_PASS }}
+    
+    - name: Release
+      run: 
+        make -f Makefile.release publish-dockerhub
+
```
rikez, line 93: nit: extra line

asaf-erlich:
So I reviewed and everything LGTM. But I wanted to chat about this a bit more:

> Test release on tag creation (once merged to master)

So this caused a significant amount of pain for @sindhusegment and also for me afterwards needing to fix the release process when we lost the creds once and no one had any idea why. Basically an annoying problem is that the release is only ever executed on tag creation which maybe makes sense but also never tests a bunch of code. 

Do you think we should be testing this release flow during merge to master for example (not necessarily on every commit) so that we find out if it's broken prior to trying a release and discovering the problem? This is obviously out of scope for this ticket? Should we have another ticket to discuss the best way to solve this? Thoughts?

.github/workflows/build.yml
```diff
@@ -0,0 +1,80 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup Go 
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go }}
+    
+    - name: Test
+      run: make test
```
asaf-erlich, line 44: Do you know what the output of make test is? Do we know what the test coverage is and whether we're satisfied with it?

asaf-erlich (approved):
Overall LGTM, comments added about things but they're all out of scope potential backlog items to discuss later.

.github/workflows/build.yml
asaf-erlich: I just looked at it in the job, but don't know the coverage, should we add a coverage report (backlog ticket)?

mckern: That looks like it's ported over 1:1 from the circleci yaml; probably something we can reasonably say "this is fine" to and sleep well tonight.

.github/workflows/release.yml
```diff
@@ -0,0 +1,93 @@
...
+        name: dist
+        path: 'dist/*'
+
+    - name: Setup Go 
+      uses: actions/setup-go@v2
+      with:
+        go-version: ${{ matrix.go }}
+
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
```
mckern, line 80: I'm assuming this is because docker is doing cross-compilation, and this is intentional (not copypasta?)

rikez: Yes, it is because of that, we're building for multiple OS. We got this from this [doc](https://github.com/marketplace/actions/build-and-push-docker-images#usage)

.github/workflows/build.yml
rikez: We've tried to port 1:1 without refactoring anything. We can improve chamber release separately.

rikez:
@asaf-erlich We should think about a better process the test the release process without having to ship to master or on tag creation. Perhaps we can release a "pre-release" version when we merge to master or a even dryrun process for every commit in a feature branch?

sindhusegment:
@asaf-erlich totally agree with you. This is something that we should definitely streamline and create a backlog ticket for

emmy-byrne-segment (approved):

emmy-byrne-segment:
LGTM. We might need to remove the circle webhook to get this to merge. The comments others had said are about what i would comment as well, so +1 to everything else but not blocking.

nice work!~

rikez:
@emmy-byrne-segment We will remove  after we ensure actions is working. Given the unknowns about CircleCI permissions, I prefer to avoid having to reconfigure circleci webhook if necessary.

PR #337, opened 2022-04-08
Automate release process using a github action module

rikez:
The [PR merge](https://github.com/segmentio/chamber/pull/335) we made created an error when publishing to github. We decided to move away from the `github-release` step in the Makefile and use Github actions steps instead to publish to github. Since this is automated and more streamlined that what the Makefile is doing. Through this, we also decided to use the automated CHANGELOG generation provided by Github Actions.

Here is the error for the PR that was merged: https://github.com/segmentio/chamber/actions/runs/2116343973

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -115,3 +115,4 @@ jobs:
         draft: true
         generate_release_notes: true
         files: dist/chamber*
+        tag_name: v2.10.10
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,3 +91,27 @@ jobs:
       with:
         username: ${{ secrets.DOCKERHUB_USER }}
         password: ${{ secrets.DOCKERHUB_PASSWD }}
+
+  test-github-release:
+    runs-on: ubuntu-latest
+    needs: dist
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


alecjacobs5401 (commented):
Changes look reasonable to me. Seems like a much simpler approach 👍

Just double checking, though, does the output paths of the files match what was previously generated in the other tool?

For example, the current release artifacts on GitHub have the pattern
```
chamber-v<VERSION>-<OS>-<ARCHITECHTURE>
```
Will this new release tool do the same format? Want to make sure there is consistency in case we have some automated tool that downloads based on that format.

Also, the example in the PR body shoes the Docker Publish failing as well. Is that still a concern?

rikez:
@alecjacobs5401 We're using the docker hub creds that belong to a different account that's why it's failing. We're waiting a response from IT to get the creds for the correct account.

alecjacobs5401 (approved):
LGTM then.

Do we have plans to remove the CircleCI webhook given that we removed the CircleCI configuration?

rikez:
@alec

> Do we have plans to remove the CircleCI webhook given that we removed the CircleCI configuration?

I'm looking forward to remove that once we ensure everything is working. Let's say we need to rollback for some reason, I wouldn't like to depend on IT to get us another webhook url.

Issue #342, opened 2022-05-04
Feature request: linux/arm64 docker builds on dockerhub


kevcube:

kevcube:
fixed in #348

PR #343, opened 2022-05-06
Docker ARM64 builds with Docker Buildx

knorby:
I ran into this [same issue](https://github.com/segmentio/chamber/issues/342), as your Docker release is only packaged for `linux/amd64`. It looks like your github actions are already setup to use docker buildx with multi-platform builds, so this PR just updates your docker build to package for `linux/arm64` as well. I can't test your github actions too easily, but hopefully this quick change just works; I verified it locally, but buildx sometimes needs a bit more setup.

### DIFF
--- a/Makefile.release
+++ b/Makefile.release
@@ -11,17 +11,15 @@ ifneq (,$(findstring -,$(VERSION)))
 endif
 
 publish-dockerhub:
-	docker build \
+	docker buildx build \
 		-t segment/chamber:$(DOCKERHUB_TAG_PREFIX)$(VERSION_MAJOR_MINOR_PATCH) \
 		-t segment/chamber:$(DOCKERHUB_TAG_PREFIX)$(VERSION_MAJOR_MINOR) \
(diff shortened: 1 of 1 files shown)


asaf-erlich:
@emmy-byrne-segment since you're more of an expert here do you know if we have any concerns with switching over to docker buildx in this PR? Would we prefer to rely on separate arm64 runners here instead for the arm64 build?

emmy-byrne-segment:
I think that it's fine to use buildx here because this is for the container versions and the executables are what need testing on different platforms.

However the dockerfile line here:

https://github.com/segmentio/chamber/blob/master/Dockerfile#L10

Is calling the makefile line here:

https://github.com/segmentio/chamber/blob/master/Makefile#L40

Which would use the amd64 executable and us publishing an an arm64 version of the container wouldn't change that.

The dockerfile would need to be changed to also build the correct excutable from the makefile based on the platform

knorby:
Closing in favor of https://github.com/segmentio/chamber/pull/348

PR #347, opened 2022-05-24
fix: enhance error message to clarify validity of forward slashes in service names

### LINKED ISSUE #333: doesn't seem to work with path based secret that starts with a '/'
for some reason i need to remove the starting slash from my chamber calls:

``` 
chamber list-services /this/is/test 
```
returns nothing

```
chamber export /this/is/test  -f yaml
Error: Failed to validate service /this/is/test: Failed to validate service name '/this/is/test'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names
```

it says it should be able to, but fails.
* if i remove the leading slash it works.

alecjacobs5401:
enhance error message to clarify validity of forward slashes in service names

Resolves #333

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -110,7 +110,7 @@ func validateService(service string) error {
 		}
 	} else {
 		if !validServicePathFormat.MatchString(service) {
-			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names", service)
+			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names. Service names must not start or end with a forwardslash", service)
 		}
 	}
 
@@ -125,7 +125,7 @@ func validateServiceWithLabel(service string) error {
 		}
 	} else {
 		if !validServicePathFormatWithLabel.MatchString(service) {
-			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names, and colon followed by a label name", service)
+			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names, and colon followed by a label name. Service names must not start or end with a forwardslash or colon", service)
 		}
 	}
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -106,11 +106,11 @@ func validateService(service string) error {
 	_, noPaths := os.LookupEnv("CHAMBER_NO_PATHS")
 	if noPaths {
 		if !validServiceFormat.MatchString(service) {
-			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, fullstops and underscores are allowed for service names", service)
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -125,7 +125,7 @@ func validateServiceWithLabel(service string) error {
 		}
 	} else {
 		if !validServicePathFormatWithLabel.MatchString(service) {
-			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names, and colon followed by a label name", service)
+			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names, and colon followed by a label name. Service names must not start or end with a forwardslash or colon", service)
```
emmy-byrne-segment, line 128: Why does this one say colon but not the above one? Because the label can have colons?

alecjacobs5401: Yep! Exactly

Difference in the regex's:
```
validServicePathFormat          = regexp.MustCompile(`^[\w\-\.]+(\/[\w\-\.]+)*$`)
validServicePathFormatWithLabel = regexp.MustCompile(`^[\w\-\.]+((\/[\w\-\.]+)+(\:[\w\-\.]+)*)?$`)
```
and the [tests show that as well](https://github.com/segmentio/chamber/blob/ade72496d6ae1391006a218a569862dd7ed07f20/cmd/root_test.go#L112-L121)

cmd/root.go
```diff
@@ -110,7 +110,7 @@ func validateService(service string) error {
 		}
 	} else {
 		if !validServicePathFormat.MatchString(service) {
-			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names", service)
+			return fmt.Errorf("Failed to validate service name '%s'.  Only alphanumeric, dashes, forwardslashes, fullstops and underscores are allowed for service names. Service names must not start or end with a forwardslash", service)
```
mckern, line 113: Nitpick: only using a single space everywhere else, and "forward slash" and "full stop" are two words each.

(I know you didn't make those typos -- just flagging them while this is open.)

```suggestion
			return fmt.Errorf("Failed to validate service name '%s'. Only alphanumeric, dashes, forward slashes, full stops and underscores are allowed for service names. Service names must not start or end with a forward slash", service)
```

cmd/root.go
mckern: Same nitpick for typos as above

```suggestion
			return fmt.Errorf("Failed to validate service name '%s'. Only alphanumeric, dashes, forward slashes, full stops and underscores are allowed for service names, and colon followed by a label name. Service names must not start or end with a forward slash or colon", service)
```

cmd/root.go
alecjacobs5401: Thanks for pointing it out! Added in https://github.com/segmentio/chamber/pull/347/commits/843d594527ac20891d5f088273c9c1630edbb38a (found some other usages of `fullstops` and double spaces)

cmd/root.go
alecjacobs5401: Thanks for pointing it out! Added in https://github.com/segmentio/chamber/pull/347/commits/843d594527ac20891d5f088273c9c1630edbb38a (found some other usages of `fullstops` and double spaces)

mckern (approved):

PR #348, opened 2022-05-31
arm64 docker builds

kevcube:
Supercedes #343

- arm64 docker builds for chamber


kevcube:
includes @knorby's work

Dockerfile
```diff
@@ -3,11 +3,12 @@ FROM golang:1.13-alpine AS build
 WORKDIR /go/src/github.com/segmentio/chamber
 COPY . .
 
+ARG TARGETARCH
```
kevcube, line 6: this is a default arg of buildkit-based builds

kevcube:
@emmy-byrne-segment this should address [this comment](https://github.com/segmentio/chamber/pull/343#issuecomment-1122715508)

Makefile
```diff
@@ -8,6 +8,10 @@ ifndef VERSION
 	VERSION := $(shell git describe --tags --always --dirty="-dev")
 endif
 
+ifndef TARGETARCH
+	TARGETARCH := $(shell arch)
```
emmy-byrne-segment, line 12: An issue we might have here is that OS X reports `arm64` but linux reports `aarch64` when running this command on the same machine and arch.

since *nix varieties cannot agree here i think we need to make sure this is going to output exactly what we expect it to.

kevcube:
@emmy-byrne-segment I just made an extra name for the linux target so that it'll catch both arm64/aarch64; is this an acceptable solution?

emmy-byrne-segment (approved):

emmy-byrne-segment:
@kevcube that should work unless i am mistaken, thank you!

Just checking on this dockerhub failure before we merge this

emmy-byrne-segment:
Turns out thats working as expected because i forgot this was a fork and that test wont work on branches. Thank you again for the PR! Merging.

Issue #349, opened 2022-06-15
chore: installation on darwin + Go 1.18


mrkagelui:
Hi, there seems to be two issues for this setup.

1. `go get` [is deprecated](https://go.dev/doc/go-get-install-deprecation) and yet in the readme and wiki, I think `go install github.com/segmentio/chamber/v2@latest` should be recommended
2. `go install` is also failing:

```command
➜  ~ go version
go version go1.18.2 darwin/amd64
➜  ~ go install github.com/segmentio/chamber/v2@latest
# golang.org/x/sys/unix
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/syscall_darwin.1_13.go:29:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.1_13.go:27:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.1_13.go:40:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.go:28:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.go:43:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.go:59:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.go:75:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.go:90:3: //go:linkname must refer to declared function or variable
go/pkg/mod/golang.org/x/sys@v0.0.0-20200930185726-fdedc70b468f/unix/zsyscall_darwin_amd64.go:105:3: //go:linkname must refer to declared function or variable
[...]

PR #350, opened 2022-06-15
chore: installation for Go 1.18

mrkagelui:
addresses #349

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
   install-go-modules:
     strategy:
       matrix:
-        go: ['1.16.x', '1.15.x']
+        go: ['1.18.x', '1.17.x', '1.16.x', '1.15.x']
     
     runs-on: ubuntu-latest
 
@@ -24,12 +24,12 @@ jobs:
         go-version: ${{ matrix.go }}
 
     - name: Test install Go modules for v${{ matrix.go }} 
-      run: go get -v . && chamber version
+      run: go install -v . && chamber version
 
   test:
     strategy:
       matrix:
-        go: ['1.16.x']
+        go: ['1.18.x', '1.17.x', '1.16.x']
     runs-on: ubuntu-latest
 
     steps:
@@ -57,7 +57,7 @@ jobs:
   dist:
     strategy:
       matrix:
-        go: ['1.16.x']
+        go: ['1.18.x', '1.17.x', '1.16.x']
     runs-on: ubuntu-latest
     needs: test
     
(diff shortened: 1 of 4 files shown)


mrkagelui:
> Would it be too much to ask you to add 1.17.x and 1.18.x to the github action build matrix here: https://github.com/segmentio/chamber/blob/master/.github/workflows/build.yml#L14

certainly, if that's ok

mrkagelui:
> Would it be too much to ask you to add 1.17.x and 1.18.x to the github action build matrix here: https://github.com/segmentio/chamber/blob/master/.github/workflows/build.yml#L14

how about `test` and `dist` though, should they be added or are they not supported?

mrkagelui:
I'll just add and ask for forgiveness 😆

asaf-erlich:
> how about test and dist though, should they be added or are they not supported?

My gut tells me to update those as well. I don't see any reason why they should fail. But I don't want to throw unnecessary work on you. Feel free to try it but if it fails don't worry about it.

mrkagelui:
> > how about test and dist though, should they be added or are they not supported?
> 
> My gut tells me to update those as well. I don't see any reason why they should fail. But I don't want to throw unnecessary work on you. Feel free to try it but if it fails don't worry about it.

done, thanks for attending to this! I'm kinda surprised no one has encountered this yet.

asaf-erlich:
Yup: https://go.dev/doc/go-get-install-deprecation:

>Starting in Go 1.17, installing executables with go get is deprecated. go install may be used instead.

>In Go 1.18, go get will no longer build packages; it will only be used to add, update, or remove dependencies in go.mod. Specifically, go get will always act as if the -d flag were enabled.

asaf-erlich:
I don't want to block your PR, you have a couple of options:

1. Just drop 1.18.x for now for the install.
2. Change it to use go install and hope that's backwards compatible for all the versions?

mrkagelui:
I don't have other go version installed locally, so here's what I'm gonna try:

1. shoot in the dark and go with `install`;
2. if that fails, I'm gonna split the job into two, >1.17 and older, wdyt?

asaf-erlich:
Alright everything passed except that dockerhub login which I think was broken before or is at least flaky, I'll approve the PR.

asaf-erlich:
cc @rikez do you know why the docker hub login github action task fails? I remember it happened before for you

rikez:
@asaf-erlich It is failing because secrets aren't available in repo forks. Basically, the job is attempting to log in to Docker Registry using empty login and password values.

This docker step should be removed from feature branches. Let me open a PR to remove it.

rikez:
https://github.com/segmentio/chamber/pull/351

mrkagelui:
@asaf-erlich would be great if we can get a new version of this so that installation is easier

mrkagelui:
> @mrkagelui I will create a backlog ticket for the team to prioritize a new release.

awesome, thank you and the team for this great tool, lots of ❤️

Issue #363, opened 2022-12-07
CVE-2022-27664 - net/http in Go before 1.18.6 and 1.19.x before 1.19.1


estokes-vs:
Our security scanner flagged `chamber` for a Go vulnerability that exists in versions older then 1.18.6 and 1.19.1.  It looks like `chamber` uses Go 1.13.  Is there a current effort to patch this vulnerability or can it be added to the roadmap?  Thanks!

https://nvd.nist.gov/vuln/detail/CVE-2022-27664

rikez:
Hi @estokes-vs, I filed a ticket internally to investigate this vuln. We will be looking into that soon.

rikez:
@estokes-vs This is mitigated on chamber [v2.11.0](https://github.com/segmentio/chamber/releases/tag/v2.11.0).

PR #364, opened 2022-12-12
feat: update go version

rikez:
Resolves https://github.com/segmentio/chamber/issues/363

Updates go version from 1.13 to 1.19.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
   install-go-modules:
     strategy:
       matrix:
-        go: ['1.18.x', '1.17.x', '1.16.x', '1.15.x']
+        go: ['1.19.x']
     
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -11,7 +11,7 @@ jobs:
   install-go-modules:
     strategy:
       matrix:
-        go: ['1.18.x', '1.17.x', '1.16.x', '1.15.x']
+        go: ['1.19.x']
```
alecjacobs5401, line 14: Does this constitute a breaking change if we have to restrict  to only 1.19+ going forward? Does this imply we're dropping all support for <1.19?

asaf-erlich:
Can you double check the brew public formula to confirm how it builds chamber and check if it's using the supported one in golang version 1.19.x? Also curious if we want to cc @mckern to gut check if there's a risk in this upgrade we aren't considering?

PR #365, opened 2022-12-13
fix: mitigate CVE-2022-27664

rikez:
Resolves https://github.com/segmentio/chamber/issues/363

Install patch that fixes vulnerability in net/http


emmy-byrne-segment (dismissed):

mckern (changes requested):
Doesn't this need to be [updated in release.yml too](https://github.com/segmentio/chamber/blob/master/.github/workflows/release.yml)?

mckern (approved):
I think this will Do The Thing. You'll need to push a tag when it's merged, and I believe the release notes for the new release will be generated for all commits between 0fda2320 and HEAD so maybe cut it as v2.11.0? I assume the commits in this PR will be squash merged, so that might be a good space to condense notes down into something that looks like release notes.

Issue #369, opened 2023-02-08
Bug Importing UPPER_CASE_KEYs


pop:
Hello 👋 when using `chamber import` with keys that are UPPER_CASE chamber is unable to read or delete those values. Here is an example:

```
$ echo '{"abc": "secretVal", "DEF": "secretVal2"}' | chamber import foo -
Successfully imported 2 secrets
$ chamber list foo
Key     Version         LastModified            User
DEF     1               2023-02-08 14:31:26     arn:aws:sts:: 111111111111:assumed-role/pop
abc     1               2023-02-08 14:31:26     arn:aws:sts:: 111111111111:assumed-role/pop
$ chamber read foo abc    
Key     Value           Version         LastModified            User
abc     secretVal       1               2023-02-08 14:31:26     arn:aws:sts:: 111111111111:assumed-role/pop
$ chamber read foo DEF
Error: Failed to read: secret not found
$ chamber delete foo DEF
Error: secret not found
$ chamber env foo     
export DEF=secretVal2
export ABC=secretVal
```

The issue is not present with manually `chamber write`-ing a secret like so:

```
$ chamber write myservice asdf supersecretvalue
$ chamber read myservice asdf
Key     Value                   Version         LastModified            User
asdf    supersecretvalue        1               2023-02-08 14:30:03     arn:aws:sts:: 111111111111:assumed-role/pop
$ chamber write myservice ASDF supersecretvalue2
$ chamber read myservice ASDF
Key     Value                   Version         LastModified            User
asdf    supersecretvalue2       2               2023-02-08 14:30:26     arn:aws:sts:: 111111111111:assumed-role/pop
$ chamber read myservice asdf                   
Key     Value                   Version         LastModified            User
asdf    supersecretvalue2       1               2023-02-08 14:30:03     arn:aws:sts:: 111111111111:assumed-role/pop
```

The secrets exist in SSM and can be read just fine with `chamber env` and added to the environment with `chamber exec`, but `read` and `delete` are broken for just imported UPPER_CASE secrets.

I am using chamber version `v2.11.1` with the SSM backend.

I am more than happy to make a PR to attempt to fix the issue if public contributions are welcome and I can get somebody to review the change.

alecjacobs5401:
Hey @pop, thanks for reporting!

Looks like this is a duplicate of a few other open issues currently https://github.com/segmentio/chamber/issues/324 https://github.com/segmentio/chamber/issues/160

pop:
[...]

Issue #324, opened 2022-01-05
Cannot delete keys, which are uppercase


Rapiernik:
Hi there, 
I inherited legacy code, which is associated with Chamber. I see there different keys, where majority is in the form blabla_blabla. But there are some keys written in uppercase, e.g. BLABLA_BLABLA. I can't read them from console or delete them. I'm getting "Error: secret not found", even though I see them in a list. Is there anything I can do with it?

yanigisawa:
FYI @Rapiernik - see this issue as well of others reporting the same issue. 

https://github.com/segmentio/chamber/issues/160

AndresElizondo:
Have you read [this comment](https://github.com/segmentio/chamber/issues/160#issuecomment-550557063)?
> For anyone else hitting this, you can remove the uppercased keys with the awscli:
> `aws ssm delete-parameters --name "{service}/{KEY}"`

This unblocked my workflow. I agree a solution inside chamber is needed though.

alecjacobs5401:
As of https://github.com/segmentio/chamber/releases/tag/v2.12.0, this can now be resolved with
```
chamber delete --exact-key
```

[Some more details in the README](https://github.com/segmentio/chamber#deleting)

This should provide the expected functionality in a non-breaking manner. In a future version, it is likely that this becomes the default behavior

mckern:
Closing; we may reopen this in the future if we determine there is more work to do in this area.

PR #370, opened 2023-02-13
feat: add options for import and delete normalization handling

alecjacobs5401:
### Description

Mitigates #160, #324, #369

Adds two forward-thinking options in lieu of implementing breaking change behavior.

* minor refactor to move some shared code logic/expectations to a `utils` package
* `chamber import`
  * added the `--normalize-keys` flag
* `chamber delete`
  * added the `--preserve-key` flag

### For the Reviewer

The first two commits are just minor refactors. It is recommended to review by each commit to reduce styling/refactor noise.

Please excuse my Prettier running rampant on `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,11 @@ you get from `chamber export`).
 > be written to the secrets backend in the format they exist in the source file.
 > In order to normalize keys on import, provide the `--normalize-keys` flag
 
+When normalizing keys, before write, the key will be be first converted to lowercase
+to match how `chamber write` handles keys.
+
+Example: `DB_HOST` will be converted to `db_host`.
+
 You can set `filepath` to `-` to instead read input from stdin.
 
 ### Deleting
--- a/README.md
+++ b/README.md
@@ -223,7 +223,7 @@ You can set `filepath` to `-` to instead read input from stdin.
 ### Deleting
 
 ```bash
-$ chamber delete [--preserve-key] service key
+$ chamber delete [--exact-key] service key
 ```
 
 `delete` provides the ability to remove a secret from chamber permanently,
@@ -233,7 +233,30 @@ secret once it has been deleted so care should be taken with this command.
 <!-- prettier-ignore -->
 > __Note__
 > By default, `delete` will normalize any provided keys. To change that behavior,
-> provide the `--preserve-key` flag to attempt to delete the raw provided key.
+> provide the `--exact-key` flag to attempt to delete the raw provided key.
+
+Example: Given the following setup,
+
+```bash
+$ chamber list service
+Key         Version                  LastModified      User
+apikey      2                        06-09 17:30:56    daniel-fuentes
+APIKEY      1                        06-09 17:30:34    daniel-fuentes
+```
+
+Calling
+
+```bash
+$ chamber delete --exact-key service APIKEY
+```
+
+will delete only `APIKEY` from the service and leave only
+
+```bash
+$ chamber list service
+Key         Version                  LastModified      User
+apikey      2                        06-09 17:30:56    daniel-fuentes
+```
 
 ### Finding
 
--- a/README.md
+++ b/README.md
@@ -1,17 +1,17 @@
 # Chamber
 
-Chamber is a tool for managing secrets.  Currently it does so by storing
+Chamber is a tool for managing secrets. Currently it does so by storing
 secrets in SSM Parameter Store, an AWS service for storing secrets.
 
 For detailed info about using chamber, read [The Right Way To Manage Secrets](https://aws.amazon.com/blogs/mt/the-right-way-to-store-secrets-using-parameter-store/)
 
 ## 2.0 Breaking Changes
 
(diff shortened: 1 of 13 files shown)


cmd/delete.go
```diff
@@ -17,17 +16,24 @@ var deleteCmd = &cobra.Command{
 	RunE:  delete,
 }
 
+var preserveKey bool
+
 func init() {
+	deleteCmd.Flags().BoolVar(&preserveKey, "preserve-key", false, "Prevent normalization of the provided key in order to delete any keys that may not have been previously normalized on write.")
```
alecjacobs5401, line 22: I'm not so sure how I feel about this being `preserve-key`.

On one hand, it makes it easy to quickly enable this behavior by doing
```
chamber delete --preserve-key
```
keeping custom behavior the "falsy" path.

On the other hand, it can be confusing for a few reasons:
* on a `delete` command, this could come across as "delete the value in the store but keep the key defined in the store" (essentially just nullify the value)
* it differs in naming/convention/standard from the `--normalize-key` added on the `import` command, so there isn't consistency for what is effectively the same behavior
  * Note, it felt bad to have this be `normalize-key` as well, but with a default of `true`

I'm open to any and all suggestions for making this better

README.md
```diff
@@ -175,43 +180,58 @@ $ chamber export [--format <format>] [--output-file <file>]  <service...>
...
 ```
 
 `import` provides the ability to import secrets from a json or yaml file (like the kind
 you get from `chamber export`).
 
+<!-- prettier-ignore -->
+> __Note__
+> By default, `import` will **not** normalize key inputs, meaning that keys will
+> be written to the secrets backend in the format they exist in the source file.
+> In order to normalize keys on import, provide the `--normalize-keys` flag
```
mckern, line 214: I'd suggest adding an example of normalized vs. "original" keys and illustrating what actually changes.

cmd/delete.go
mckern: How do you feel about just reusing `--normalize-key` for `delete` as well?

README.md
alecjacobs5401: Sounds good! I was relying on tests for this, but will add an example 👍

cmd/delete.go
alecjacobs5401: @mckern definitely agree with you, but removing feels like a breaking change in behavior we would have to account for in a major version bump, right?

README.md
alecjacobs5401: Added in https://github.com/segmentio/chamber/pull/370/commits/6174a82a5fca3cee236162b365d894dcab3c7014

cmd/delete.go
alecjacobs5401: Just noticed your edit - happy to move to `--no-case` over `--preserve-key` 👍

alecjacobs5401: context: we're digging into how AWS handles same keys with different casings to understand impact of deleting a key that is `MY_SECRET_VALUE` if the provided key is `my_secret_value`

alecjacobs5401: Resolved offline - AWS allows for case sensitive unique keys, so `MY_SECRET_VALUE` and `my_secret_value` are both treated as unique. Changing `--preserve-key` to `--exact-key` to make it clearer in the command line

utils/utils.go
```diff
@@ -0,0 +1,13 @@
...
+
+// NormalizeService normalizes a provided service to a common format
+func NormalizeService(service string) string {
+	return strings.ToLower(service)
+}
+
+// NormalizeKey normalizes a provided secret key to a common format
+func NormalizeKey(key string) string {
+	return strings.ToLower(key)
+}
```
mckern, line 13: What's the intent in moving these to a separate package? Since these both just call `strings.ToLower()`, right now it's just indirection on top of stdlib.

alecjacobs5401: just standardization/normalization. There are separate packages in this repo that rely on the same logic, so was attempting to DRY it out by giving it an interface

mckern (approved):
I'm not going to block on the question I left but I did **have** a question.

utils/utils.go
mckern: 🤷🏻‍♂️ I'd argue that 2 new methods that both lowercase a string is a worse deal than just using `strings.ToLower()` when needed buuuuuuut it's not a hill I care about dying on.

alecjacobs5401: Totally fair - I might be overthinking the need to update how or what normalization means in the future and implemented this as a measure.

I may have also thought that the lack of a normalizing function (and just raw `strings.ToLower`) may have results in missed normalization in the past, which lead me to move towards it

mckern: Yeah, I can see that; if any of the other normalizion/sanitization happened in these functions, I think it'd make more sense to have them. As they're written right now, they're just 1:1 wrappers around `strings.ToLower()` without any benefit for the indirection.

PR #371, opened 2023-02-21
chore: Update go version in go.mod

jmeridth:
This will match go 1.19.x that is used in the Dockerfile already

Notes:

- Why are there two require sections now?
  - Because in Go 1.17 the module graph has been changed to enable pruning and lazy loading. The second require block contains indirect dependencies
- CI
  - Lowest chamber builds for is 1.15 [here](https://github.com/segmentio/chamber/blob/301d5feb8dccca57c524ca2fda418f8aa96fcc9d/.github/workflows/build.yml#L14)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,20 +2,25 @@ module github.com/segmentio/chamber/v2
 
 require (
 	github.com/aws/aws-sdk-go v1.38.22
-	github.com/bmizerany/assert v0.0.0-20160611221934-b7ed37b82869 // indirect
-	github.com/davecgh/go-spew v1.1.1 // indirect
-	github.com/inconshreveable/mousetrap v1.0.0 // indirect
-	github.com/kr/pretty v0.1.0 // indirect
 	github.com/magiconair/properties v1.8.0
 	github.com/pkg/errors v0.9.1
-	github.com/segmentio/backo-go v0.0.0-20160424052352-204274ad699c // indirect
 	github.com/spf13/cobra v0.0.3
-	github.com/spf13/pflag v1.0.2 // indirect
 	github.com/stretchr/testify v1.2.2
-	github.com/xtgo/uuid v0.0.0-20140804021211-a0b114877d4c // indirect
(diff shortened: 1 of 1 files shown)


jmeridth:
Just noticed the circleci config.  Will add changes.

jmeridth:
@alecjacobs5401 are you a good person to take a look at this one or is someone else better?  Thank you for any insight.  Cheers.

alecjacobs5401:
Hey @jmeridth - sorry for the late response! We've been discussing this internally for a little and are thinking through potential consequences of this change immediately.

On one hand, this is a good change and we'd like to enforce a newer go version. On the other, we are investigating the potential downstream impact to implementing libraries that might pull in `chamber` packages (i.e. not just using `chamber` as a CLI, using its internals in other go packages)

We did some looking around and it looks like [mostly forks of chamber exist](https://github.com/search?q=import+%22github.com%2Fsegmentio%2Fchamber&type=code), but there is at least one Repository where `chamber` is pulled in as a package.

By changing the go version in `go.mod`, we do cause some potential for breaking behavior by now requiring any implementing packages to adhere to a new minimum go version.

I'm going to rope in @knksmith57 to handoff to since he is our current on-call engineer and we'll work out next steps.

jmeridth:
@alecjacobs5401 That makes sense.  anytime we upgrade to a minor (especially major) version, those are the things you have to triple check.  Thank you for the response.  Looking forward to seeing how this pans out.  Cheers.

knksmith57:
Hey @jmeridth I synced with the team today and I think this warrants more internal discussion based on potential blast radius.

I'm going to sync with one of our Principal's tomorrow about next steps. For now, agreed, let's close this one so it isn't floating around.

Thank you, again, for bringing this to our attention and for the PR 🙏

jmeridth:
@knksmith57 any chance of progress on this idea internally? Thank you in advance for any updates. Cheers.

jmeridth:
Solved by https://github.com/segmentio/chamber/pull/383

PR #340, opened 2022-04-28
enable importing dotenv files

kevcube:
possible solution to #339

Issue #339, opened 2022-04-27
Feature request: `import` to support .env/dotenv files


kevcube:
Similar to #18 

We use next.js, which depends on .env files.

Thankfully we are able to export in this format, but it would simplify things if we could also import in dotenv format.

knksmith57:
hey there Kevin, thanks for the suggestion!

while we wait for more input on your feature request, in case you were looking for a workaround, have you considered using a combination of serializing dotenv's output + chamber's stdin import?

I believe this will work:

```sh
node -p 'JSON.stringify(require("dotenv").config().parsed)' | chamber import -
```

explanation: `chamber import` understands yaml and JSON, and dotenv returns a JSON-serializable POJO of the parsed `.env` file.

more complete example:
```sh
## write some dummy values to .env to exercise parser edge cases
❯ > .env <<EOT
basic="foo"
multiline="my
...
multiline value
..."
comment="should be included" # should not be included
nocomment="is not # a comment"
EOT

## use node -p to print out a JSON-serialized representation of .env
❯ node -p 'JSON.stringify(require("dotenv").config().parsed)'
{"basic":"foo","multiline":"my\n...\nmultiline value\n...","comment":"should be included","nocomment":"is not # a comment"}

## or, pretty printed via jq for readability
❯ node -p 'JSON.stringify(require("dotenv").config().parsed)' | jq
{
  "basic": "foo",
  "multiline": "my\n...\nmultiline value\n...",
  "comment": "should be included",
  "nocomment": "is not # a comment"
}

## feed .env variables serialized to JSON into chamber via stdout => stdin
❯ node -p 'JSON.stringify(require("dotenv").config().parsed)' | chamber import -
```

kevcube:
@knksmith57 thanks a bunch for the detailed workaround, yes I was aware of this as an option and I'm glad you basically coded it for me 😄 

I was looking at the chamber's source code thinking "now ... WHERE do they process the Json?" and scratching my head until it became clear that the yaml library is capable of handling JSON and YAML.

I wrote some quick-and-dirty code that can handle this, tested and working for me, and saw the last in your contributor's guidelines that says not to feature PR without discussion first..

Well I already wrote the code and just wanna put it up for posterity but I'm open to any and all feedback/suggestions ❤

kevcube:
@knksmith57 when you say "while we wait for more input on your feature request", are you waiting for input from me?

knksmith57:
[...]

PR #372, opened 2023-03-28
enable importing dotenv files

mckern:
This PR Updates & supersedes #340, which was opened as a possible fix for #339
----

This introduces a dependency on https://github.com/joho/godotenv/, which itself carries no dependencies outside of Go's stdlib. I've opened this PR to address merge conflicts, as well as to begin reassessing the solution; I am convinced of the usefulness of the feature but not necessarily the ergonomics of the implementation.

My concerns/thoughts as I work through this are:

- `.env` files are not standardized, so there are many conflicting, almost-but-not-quite identical parser implementations
  + what is considered "valid" in [godotenv](https://github.com/joho/godotenv/) may not work if fed to Node's [dotenv](https://www.npmjs.com/package/dotenv) or the alternative [dotenvy](https://www.npmjs.com/package/dotenvy), or it may not be parsable by a standard compatible shell
  + It looks like `joho/godotenv` strives to adhere as close as possible to [Ruby's dotenv](https://github.com/bkeepers/dotenv) and [Node.js' dotenv](https://github.com/motdotla/dotenv) implementations. This is probably a reasonable amount of _de facto_ standardization?
- Most implementations of `.env` parsers work by treating these files as local overrides to an existing environment; they converge on a complete environment by applying the contents of a `.env` file to their inherited environment. Parsing them as key-value pairs is probably alright ?
- How much shell grammar is a `.env` file expected to support?
- How is this functionally any different than support `.ini`/`.properties`?

### DIFF
--- a/cmd/import.go
+++ b/cmd/import.go
@@ -5,6 +5,7 @@ import (
 	"io"
 	"os"
 
+	"github.com/joho/godotenv"
 	"github.com/pkg/errors"
 	"github.com/segmentio/chamber/v2/store"
 	"github.com/segmentio/chamber/v2/utils"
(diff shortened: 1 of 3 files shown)


mckern:
OK. I have worked through this, and my questions about it. Unfortunately, I don't believe this is something we should merge. I'll bring feedback back to #339 and #340 about why.

PR #379, opened 2023-05-03
Update Dockerfile to use go 1.20.4-alpine

m-graf:


martinr23 (approved):

mckern:
I don't have any particular problems with this but it's worth pointing out that this explicitly moves the base image from a floating Go revision (1.x) to a specific Go revision (1.x.x) and we (the maintainers) should be aware of that. I generally prefer specific versions for producing release artifacts, so I'm 👍🏻 on this but I want to make sure that doesn't go unnoticed by other maintainers.

mckern:
FWIW, this should be merged **after** #382. That adds Go 1.20.x into the test matrix. It may require merging `master` back into this branch, but these are reasonable changes.

mckern (approved):

PR #384, opened 2023-05-10
Fix accumulated papercuts

mckern:
images (2)

So, `chamber` has accumulated a number of papercuts over the years: small things, easy to overlook but that sting in the moment. Nothing that's going to completely halt operation, but will annoy you (read: me, Joe Sportsbar Average User) when you touch them. Here's a swing at fixing a few of these up in the hopes of further improving future maintenance: 

- Improve test readability for `cmd/find_test.go`
- Use contemporary `go:build` directives format instead of deprecated `+build:`
- Remove legacy 3rd party error wrapping package and use contemporary stdlib error wrapping

### DIFF
--- a/cmd/delete.go
+++ b/cmd/delete.go
@@ -1,11 +1,12 @@
 package cmd
 
 import (
-	"github.com/pkg/errors"
+	"fmt"
+
+	analytics "github.com/segmentio/analytics-go/v3"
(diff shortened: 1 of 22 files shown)


alecjacobs5401 (approved):
thank you for this cleanup!

Issue #366, opened 2022-12-30
`write` allows for invalid shell variable creation


jmhodges-color:
This is with chamber v2.11.0 on macOS 13.1

I think chamber should be more strict about what it accepts in names at `write` time. For instance, accepting dashes (`-`) in variable names due to two related, but interestingly different bugs described below in `env` and `export -f dotenv`.

Accepting periods (`.`) also causes invalid env var names to be created in both `env` and `export -f dotenv`, but those are not described here.

Both suppose you have written a secret to chamber with a dash in it:

```
chamber write mysvc foo-bar 1
```

### env bug
if run `chamber env mysvc`, you'll get the invalid variable `FOO-BAR` in the output. The dash in it is not supported in shell variables in both `bash` and `zsh`.

```
$ chamber env mysvc
export FOO-BAR=1

$ source <(chamber env mysvc)
/dev/fd/11:export:1: not valid in this context: FOO-BAR

```

### export dotenv bug

If you run a second write with an underscore version of the parameter, then run export with the `dotenv` format, you'll get duplicate env vars with no warning. (Interestingly, the export dotenv code path does normalize the dash to an underscore in this case)

```
$ chamber write mysvc foo_bar 2
$ chamber export -f dotenv mysvc
FOO_BAR=2
FOO_BAR=1
$ 
```

### Conclusion
We're planning on telling folks to not use dashes from variable names, but would like to have some support from `chamber` to prevent these kinds of problems at time of `write`.

alecjacobs5401:
Thanks for reporting! 

>if run `chamber env mysvc`, you'll get the invalid variable `FOO-BAR` in the output. The dash in it is not supported in shell variables in both `bash` and `zsh`.

This feels like the core bug here and I believe there are some other issues that relate. This might warrant a separate issue

> If you run a second write with an underscore version of the parameter, then run export with the `dotenv` format, you'll get duplicate env vars with no warning.

This could be related to lack of validation testing on duplicate keys where case insensitivity/normalization is not respected, which leads to duplicate keys that effectively resolve to the same environment variable (related https://github.com/segmentio/chamber/pull/162, https://github.com/segmentio/chamber/issues/160)

Due to the breaking nature of the latter problem, it would likely need to be part of a 3.0 release.

I'll bring this up with my team to prioritize looking into

mckern:
[...]

PR #385, opened 2023-05-10
Fix `chamber env` and `chamber export -f dotenv`

mckern:
This was reported in https://github.com/segmentio/chamber/issues/366: `chamber` made no effort to sanitize keys for either of these commands or formats, which can result in invalid variable names being used for both of these output formats. This PR fixes half of the problem (invalid characters in retrieved keys when rendered via `env` or `export -f dotenv`) but does not address the larger question of "what is a valid character when writing a key?"

Some of the construction of these formats have now been unified, with an optional boolean flag for `env` to preserve key case -- note that keys will still be sanitized, they just won't be converted to uppercase. Since `export -f dotenv` rewrites values using escape characters, I've made the choice to continue allowing `export -f dotenv` to always convert keys to uppercase.

## New Flags

`chamber env` has received a couple of new flags: `-e, --escape-strings`, `-p, --preserve-case`   :

- `--escape-strings` will allow `chamber env` to print values as either safely quoted string literals or safely quoted escape sequences.
- `--preserve-case` will force `chamber env` to respect whatever case a key originally used

### Before

```
$ chamber env -b s3 --backend-s3-bucket dummy-bucket example-service
export EXAMPLE_PASSWORD=cafeface1337
export EXAMPLE-ACCESS-TOKEN=abcdefghijklmnopqrstuvwxyz
export EXAMPLE-USER=itsamemario
export EXAMPLE-MULTILINE='I have eaten
the plums
that were in
the icebox

and which
you were probably
saving
for breakfast

Forgive me
they were delicious
so sweet
and so cold
'

$ chamber export -f dotenv -b s3 --backend-s3-bucket dummy-bucket example-service
EXAMPLE_PASSWORD="cafeface1337"
EXAMPLE-ACCESS-TOKEN="abcdefghijklmnopqrstuvwxyz"
EXAMPLE-USER="itsamemario"
EXAMPLE-MULTILINE="I have eaten\nthe plums\nthat were in\nthe icebox\n\nand which\nyou were probably\nsaving\nfor breakfast\n\nForgive me\nthey were delicious\nso sweet\nand so cold\n'
```

### After

```
$ chamber env -b s3 --backend-s3-bucket dummy-bucket example-service
export EXAMPLE_PASSWORD=cafeface1337
export EXAMPLE_ACCESS-TOKEN=abcdefghijklmnopqrstuvwxyz
export EXAMPLE_USER=itsamemario
export EXAMPLE_MULTILINE='I have eaten
the plums
that were in
the icebox

and which
you were probably
saving
for breakfast

Forgive me
they were delicious
so sweet
and so cold
'

$ chamber export -f dotenv -b s3 --backend-s3-bucket dummy-bucket example-service
EXAMPLE_PASSWORD="cafeface1337"
EXAMPLE_ACCESS-TOKEN="abcdefghijklmnopqrstuvwxyz"
EXAMPLE_USER="itsamemario"
EXAMPLE_MULTILINE="I have eaten\nthe plums\nthat were in\nthe icebox\n\nand which\nyou were probably\nsaving\nfor breakfast\n\nForgive me\nthey were delicious\nso sweet\nand so cold\n'
```

## What about backwards compatibility?

ce360a34866ff26bd94c2057a9a9b328

Effort has been taken to preserve the _spirit_ of backwards compatibility with the output formats that these commands have rendered in the past. But since the formats were objectively incorrect, I'm really not worried about it.

### DIFF
--- a/cmd/env.go
+++ b/cmd/env.go
@@ -3,12 +3,18 @@ package cmd
 import (
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
+	"github.com/alessio/shellescape"
 	analytics "github.com/segmentio/analytics-go/v3"
 	"github.com/segmentio/chamber/v2/utils"
+
 	"github.com/spf13/cobra"
 )
+
+// originally ported from github.com/joho/godotenv
+const doubleQuoteSpecialChars = "\\\n\r\"!$`"
 
 var (
 	// envCmd represents the env command
@@ -18,27 +24,53 @@ var (
 		Args:  cobra.ExactArgs(1),
 		RunE:  env,
 	}
-	pattern *regexp.Regexp
+	preserveCase   bool
+	escapeSpecials bool
 )
 
 func init() {
+	envCmd.Flags().BoolVarP(&preserveCase, "preserve-case", "p", false, "preserve variable name case")
+	envCmd.Flags().BoolVarP(&escapeSpecials, "escape-strings", "e", false, "escape special characters in values")
 	RootCmd.AddCommand(envCmd)
-	pattern = regexp.MustCompile(`[^\w@%+=:,./-]`)
 }
 
+// Print all secrets to standard out as valid shell key-value
+// pairs or return an error if secrets cannot be safely
+// represented as shell words.
 func env(cmd *cobra.Command, args []string) error {
+	envVars, err := exportEnv(cmd, args)
+	if err != nil {
+		return err
+	}
+
+	for i := range envVars {
+		fmt.Println(envVars[i])
+	}
+
+	return nil
+}
+
+// Handle the actual work of retrieving and validating secrets.
+// Returns a []string, with each string being a `key=value` pair,
+// and returns any errors encountered along the way.
+// Keys will be converted into valid shell variable names,
+// and converted to uppercase unless --preserve is passed.
+// Key ordering is non-deterministic and unstable, as returned
+// value from a given secret store is non-deterministic and unstable.
+func exportEnv(cmd *cobra.Command, args []string) ([]string, error) {
 	service := utils.NormalizeService(args[0])
 	if err := validateService(service); err != nil {
-		return fmt.Errorf("Failed to validate service: %w", err)
+		return nil, fmt.Errorf("Failed to validate service: %w", err)
 	}
 
 	secretStore, err := getSecretStore()
 	if err != nil {
-		return fmt.Errorf("Failed to get secret store: %w", err)
+		return nil, fmt.Errorf("Failed to get secret store: %w", err)
 	}
+
 	rawSecrets, err := secretStore.ListRaw(service)
 	if err != nil {
-		return fmt.Errorf("Failed to list store contents: %w", err)
+		return nil, fmt.Errorf("Failed to list store contents: %w", err)
 	}
 
 	if analyticsEnabled && analyticsClient != nil {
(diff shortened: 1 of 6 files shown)


cmd/env.go
```diff
@@ -53,24 +86,113 @@ func env(cmd *cobra.Command, args []string) error {
...
+
+	return out, nil
+}
+
+// The name of a variable can contain only letters (a-z, case insensitive),
+// numbers (0-9) or the underscore character (_). It may only begin with
+// a letter or an underscore.
+func validateShellName(s string) error {
+	shellChars := regexp.MustCompile(`^[A-Za-z0-9_]+$`).MatchString
+	validShellName := regexp.MustCompile(`^[A-Za-z_]{1}`).MatchString
```
alecjacobs5401, line 146: for these regexps, would it make sense to add them to our [list of valid format regexps](https://github.com/segmentio/chamber/blob/e032a23a8c16f4bd86f0da03bdbf2127dfc89717/cmd/root.go#L19-L23)?

cmd/env.go
```diff
@@ -53,24 +86,113 @@ func env(cmd *cobra.Command, args []string) error {
...
+			toReplace = `\r`
+		}
+		line = strings.Replace(line, string(c), toReplace, -1)
+	}
+	return line
+}
+
+func sortedKeys(params map[string]string) []string {
+	keys := make([]string, len(params))
+	i := 0
```
alecjacobs5401, line 191: What is this doing? I just see it get incremented down below but not actually used anywhere

cmd/env_test.go
```diff
@@ -0,0 +1,54 @@
...
+	}{
+		{givenName: "invalid strings", expectedName: "invalid_strings"},
+		{givenName: "extremely  invalid  strings", expectedName: "extremely__invalid__strings"},
+		{givenName: fmt.Sprintf("\nunbelievably\tinvalid\tstrings\n"), expectedName: "unbelievably_invalid_strings"},
+		{givenName: "valid_string", expectedName: "valid_string"},
+		{givenName: "validish-string", expectedName: "validish_string"},
+		// these strings should not be corrected, simply returned as-is
+		{givenName: "1invalidstring", expectedName: "1invalidstring"},
+		{givenName: "_1validstring", expectedName: "_1validstring"},
+		{givenName: "invalid.string", expectedName: "invalid_string"},
```
alecjacobs5401, line 44: this does not adhere to the comment above saying

> // these strings should not be corrected, simply returned as-is

is that expected?

alecjacobs5401 (commented):
Nice.

Nothing immediately glaring at me, but a few open questions.

Thanks for doing this @mckern! Having this be consistent it amazing

cmd/env_test.go
mckern: might be copy-pasta from an older iteration of the test

cmd/env.go
mckern: check again -- line 193. it's used as the index to assign values deterministically to `keys`. that being said, this function was moved here wholesale from export.go, with no changes. it's probably better expressed as a 3-part loop.

alecjacobs5401: 🤦 
I read the loop body like 3 times and missed it.

Thanks.

cmd/env_test.go
```diff
@@ -0,0 +1,54 @@
...
+func Test_validateShellName(t *testing.T) {
+	tests := []struct {
+		name       string
+		str        string
+		shouldFail bool
+	}{
+		{name: "strings with spaces should fail", str: "invalid strings", shouldFail: true},
+		{name: "strings with only underscores should pass", str: "valid_string", shouldFail: false},
+		{name: "strings with dashes should fail", str: "validish-string", shouldFail: true},
+		{name: "strings that start with numbers should fail", str: "1invalidstring", shouldFail: true},
```
alecjacobs5401, line 17: do we need a strings that include a number (but don't start with) should pass test?

cmd/env_test.go
```diff
@@ -0,0 +1,54 @@
...
+		expected string
+	}{
+		{given: "invalid strings", expected: "invalid_strings"},
+		{given: "extremely  invalid  strings", expected: "extremely__invalid__strings"},
+		{given: fmt.Sprintf("\nunbelievably\tinvalid\tstrings\n"), expected: "unbelievably_invalid_strings"},
+		{given: "valid_string", expected: "valid_string"},
+		{given: "validish-string", expected: "validish_string"},
+		{given: "valid.string", expected: "valid_string"},
+		// the following two strings should not be corrected, simply returned as-is.
+		{given: "1invalidstring", expected: "1invalidstring"},
```
alecjacobs5401, line 43: similar question on just string including a number

alecjacobs5401 (approved):
I have really one comment repeated about a test case for when we're validating/sanitizing shell keys.

I have no strong feelings one way or another and this LGTM as is.

Thanks!

cmd/env_test.go
mckern: eh, feels redundant; it's a valid character -- what matters is it **doesn't start** with a number so that's (I believe) the only case we actually need to screen for.

cmd/env_test.go
mckern: similar answer!

PR #386, opened 2023-05-16
Update `README` for pending 2.13.0 release

mckern:
Adds information about which versions of Go we will test against and support, which version of Go we will build distributable executables with, caveats about `chamber env` and `chamber export -o dotenv`, and general formatting tweaks to make `markdownlint` yell at me a little less while I was in here.


alecjacobs5401 (dismissed):
LGTM, just one minor comment about a missing space, no hard requirements here

alecjacobs5401 (approved):
🚀

PR #404, opened 2023-06-30
docs: Document the list-services command

dguo:
I was looking for a way to list all the services, and I found https://github.com/segmentio/chamber/issues/141 and https://github.com/segmentio/chamber/pull/187, but I noticed that the README doesn't mention the command.


alecjacobs5401 (approved):
💯
Thanks for taking a moment to improve our documentation! Much appreciated!

PR #415, opened 2023-08-24
store: fail faster if not running in EC2

kevinburkesegment:
If you are running `chamber` outside of EC2 without a region configured, the binary will take 5 seconds to fail because the default ec2 metadata API timeout is 5 seconds.

Instead, take a few milliseconds to determine whether we are even running on EC2. We attempt to do this by checking the contents of different files on the disk, e.g. /sys/devices/virtual/dmi/id/board_asset_tag, and then if all of those fail, attempting to see if anything is listening on the metadata host.

This can help prevent the 5 second failure timeout.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.19
 require (
 	github.com/alessio/shellescape v1.4.1
 	github.com/aws/aws-sdk-go v1.44.312
+	github.com/kevinburke/isec2 v0.1.0
 	github.com/magiconair/properties v1.8.7
 	github.com/segmentio/analytics-go/v3 v3.2.1
 	github.com/spf13/cobra v1.7.0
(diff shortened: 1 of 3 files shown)


mckern:
What's the advantage to using this approach instead of using `ec2metadata`'s [`Available()`](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.44.331/aws/ec2metadata#EC2Metadata.Available)?

It looks like `Available()` is intended for exactly this use case, and implementing `Available()` is a very small addition:

```diff
 		ec2metadataSvc := ec2metadata.New(session)
+		if !ec2metadataSvc.Available() {
+			return nil, nil, err
+		}
+
```

kevinburkesegment:
Reading through the source I believe that that endpoint would also have a 5 second timeout. I can test it experimentally though.

mckern:
> Reading through the source I believe that that endpoint would also have a 5 second timeout. I can test it experimentally though.

Yeah, I think you're right but can't we just feed it a temporary timeout context (note, this is **decidedly?** incomplete, and only serves as an example):

```diff
 		ec2metadataSvc := ec2metadata.New(session)
+		ctx, cancelFn := context.WithTimeout(aws.BackgroundContext(), 20*time.Millisecond)
+		defer cancelFn()
+		if !ec2metadataSvc.AvailableWithContext(ctx) {
+			return nil, nil, fmt.Errorf("EC2 Metadata service not available")
+		}
+
```

```
chamber $ aws-okta exec ci-read -- time ./test.sh
+ unset AWS_DEFAULT_REGION AWS_REGION
+ chamber --backend s3 --backend-s3-bucket segment-secrets-ci export whatever
Error: EC2 Metadata service not available
        0.08 real         0.03 user         0.01 sys
```

nijave:
I don't think you can rely on `/sys` for determining if you're on EC2. You could be running in a jail/chroot or container and not necessarily have `/sys` access but be able to access the metadata endpoint.

It might be more reliable to add a flag to disable EC2 metadata check

PR #426, opened 2023-10-10
.github: update to Go 1.21

kevinburkesegment:
Also update go.mod version and dependencies to latest versions.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,12 +10,12 @@ jobs:
   install-go-modules:
     strategy:
       matrix:
-        go: ["1.20.x", "1.19.x"]
+        go: ["1.21.x", "1.20.x", "1.19.x"]
 
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
 
       - name: Setup Go ${{ matrix.go }}
         uses: actions/setup-go@v4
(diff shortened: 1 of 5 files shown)


.github/workflows/build.yml
```diff
@@ -28,11 +28,11 @@ jobs:
   test:
     strategy:
       matrix:
-        go: ["1.20.x", "1.19.x"]
+        go: ["1.21.x", "1.20.x"]
```
alecjacobs5401, line 31: are we missing `1.19.x` here as well? Looks like we set that for installing modules and for the `dist` strategy below

alecjacobs5401 (changes requested):
One minor comment

Also, can we make sure the [Dockerfile value](https://github.com/segmentio/chamber/blob/master/Dockerfile#L1) matches as well?

Thanks for opening this up!

go.mod
```diff
@@ -1,22 +1,22 @@
 module github.com/segmentio/chamber/v2
 
-go 1.19
```
alecjacobs5401, line 3: do we need to bump this as well?

IIRC, we hesitated on bumping this because of breaks to packages that pull in `chamber` as a dependency.

I would vote to keep this on 1.19 now since this represents a _minimum_ version for pulling in

kevinburkesegment: Ah, you're right

.github/workflows/build.yml
kevinburkesegment: This is fixed

go.mod
kevinburkesegment: This is fixed

knksmith57 (approved):

alecjacobs5401 (approved):

PR #453, opened 2024-01-16
ci: bump actions/upload-artifact and actions/download-artifact to v4

alecjacobs5401:
### Description

We're seeing some consistent 400 responses in our `dist` steps for our GitHub actions. After doing some digging, it looks like the [v4 version of the Artifacts GitHub Actions](https://github.blog/changelog/2023-12-14-github-actions-artifacts-v4-is-now-generally-available/) was made available in early December that should resolve and/or make more reliable these actions.

### Changes

The `v4` versions of these actions come with some breaking changes and requirements that are addressed in this PR:
* If we are using `actions/upload-artifacts@v4`, we _**must also**_ be using `actions/download-artifacts@v4`
* Artifact uploads **must** contain unique names.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: Make distributables
         run: make -f Makefile.release dist
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: dist
           path: "dist/*"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -22,7 +22,7 @@ jobs:
 
       - name: Make distributables
         run: make -f Makefile.release dist
-      - uses: actions/upload-artifact@v3
+      - uses: actions/upload-artifact@v4
         with:
           name: dist
```
alecjacobs5401, line 27: I opted to leave this as `dist`, but was tempted to make it `dist-release` for more defined uniqueness. Any input?

Sabrina0614 (approved):

PR #468, opened 2024-03-08
build: update to Go 1.22

bhavanki:
Includes:

* Bumping go.mod to minimum Go 1.20
* Updating GitHub build and release workflows to 1.22, and dropping 1.19
* Bumping the Docker base image to Go 1.22.1

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,7 +10,7 @@ jobs:
   install-go-modules:
     strategy:
       matrix:
-        go: ["1.21.x", "1.20.x", "1.19.x"]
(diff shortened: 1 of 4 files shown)


alecjacobs5401 (approved):
LGTM!

[I did a quick search](https://github.com/search?type=code&q=path%3A**%2Fgo.mod+segmentio%2Fchamber++NOT+is%3Aarchived+NOT+is%3Afork) and there doesn't seem to be anything public that is consuming `chamber` as a package that wasn't already on an unsupported go version.

Thanks for making this update!

PR #481, opened 2024-04-23
chore: Upgrade to aws-sdk-go-v2

bhavanki:
This is a major update to chamber's support for the S3, SSM, and Secrets Manager store implementations. Every effort was made to preserve functionality, but there is one gap.

The v2 SDK does not expose a retryer field for a minimum throttle delay, so that argument is currently ignored when constructing new SSM stores. Support for the delay will be addressed later.

The v2 SDK does not offer "iface" interfaces for the various clients, so instead interfaces tailored to what chamber uses are defined. For testing, these new interfaces are mocked, and mock types are generated using github.com/matryer/moq. You don't need moq to use chamber or even to build it, but only if you are developing chamber and make a change to an API interface.

Also, old code in the SSM store implementation that allowed it to work without IAM permissions for ssm:GetParametersByPath has been eliminated. The permissions have been expected for a long time now.


erikdw:
FWIW, I've been running with this version of chamber for a week and I use it pretty often.

store/s3store.go
```diff
@@ -134,12 +129,12 @@ func (s *S3Store) Write(id SecretId, value string) error {
...
 	putObjectInput := &s3.PutObjectInput{
 		Bucket:               aws.String(s.bucket),
-		ServerSideEncryption: aws.String(s3.ServerSideEncryptionAes256),
+		ServerSideEncryption: types.ServerSideEncryptionAes256,
 		Key:                  aws.String(objPath),
 		Body:                 bytes.NewReader(contents),
 	}
 
-	_, err = s.svc.PutObject(putObjectInput)
+	_, err = s.svc.PutObject(context.TODO(), putObjectInput)
```
asaf-erlich, line 137: Shouldn't the same context be passed down by the CLI to each of these methods? Sorry if I'm incorrect with my assumptions here about the way context is supposed to work.

asaf-erlich: Does it work? By that I mean if you press Ctrl+C while it's doing something does it properly handle it / exit?

store/ssmstore.go
```diff
@@ -265,28 +273,33 @@ func (s *SSMStore) ListServices(service string, includeSecretName bool) ([]strin
 
 	if s.usePaths {
 		describeParametersInput = &ssm.DescribeParametersInput{
-			MaxResults: aws.Int64(50),
-			ParameterFilters: []*ssm.ParameterStringFilter{
+			MaxResults: aws.Int32(50),
```
mckern, line 276: functional change or stylistic change?

mckern (commented):
Gonna build this and put it through some functional tests but nothing jumps out at me.

mckern:
What're you thinking about for a version number on this? Clearly not a patch release but I'm not sure it's a major. Minor?

store/s3store.go
asaf-erlich: If this wasn't something that worked before it's fine to leave it as is.

Makefile
```diff
@@ -32,7 +43,7 @@ dist/:
 
 build: chamber
 
-chamber:
+chamber: $(SRC)
```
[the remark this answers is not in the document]
bhavanki: That's the idea. Notably, without this, chamber would always be rebuilt even if no .go file changed. I'm not wedded to this change, so if you have concerns, I'm happy to drop it.

store/awsapi.go
```diff
@@ -0,0 +1,47 @@
...
+	GetCallerIdentity(ctx context.Context, params *sts.GetCallerIdentityInput, optFns ...func(*sts.Options)) (*sts.GetCallerIdentityOutput, error)
+}
+
+type apiSecretsManager interface {
+	CreateSecret(ctx context.Context, params *secretsmanager.CreateSecretInput, optFns ...func(*secretsmanager.Options)) (*secretsmanager.CreateSecretOutput, error)
+	DescribeSecret(ctx context.Context, params *secretsmanager.DescribeSecretInput, optFns ...func(*secretsmanager.Options)) (*secretsmanager.DescribeSecretOutput, error)
+	GetSecretValue(ctx context.Context, params *secretsmanager.GetSecretValueInput, optFns ...func(*secretsmanager.Options)) (*secretsmanager.GetSecretValueOutput, error)
+	ListSecretVersionIds(ctx context.Context, params *secretsmanager.ListSecretVersionIdsInput, optFns ...func(*secretsmanager.Options)) (*secretsmanager.ListSecretVersionIdsOutput, error)
+	PutSecretValue(ctx context.Context, params *secretsmanager.PutSecretValueInput, optFns ...func(*secretsmanager.Options)) (*secretsmanager.PutSecretValueOutput, error)
+}
```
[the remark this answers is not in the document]
bhavanki: Indeed it does. Adding!

store/ssmstore.go
bhavanki: The v2 SDK changes the field type.

bhavanki:
> What're you thinking about for a version number on this? Clearly not a patch release but I'm not sure it's a major. Minor?

At least minor, for sure. But, I was thinking we could use this disruption as an opportunity to remove some deprecated features. For example, in this commit I removed some old workaround code for a IAM permission that became required in the distant past. Those removals would count as breakage, meriting a major release.

asaf-erlich (approved):

mckern:
> Those removals would count as breakage, meriting a major release.

Ah, if a major version is on the table then do we want to talk about what else might go into it and hold off releasing after this is merged?

bhavanki:
> > Those removals would count as breakage, meriting a major release.
> 
> Ah, if a major version is on the table then do we want to talk about what else might go into it and hold off releasing after this is merged?

Definitely! There's at least one other thing I recall spotting that could be dropped, I forget what it is at the moment. Also, there are some improvements that can follow from this PR, like how contexts are handled. I don't see any problem letting this update sit unreleased on master while we do some spring cleaning.

mckern (approved):

PR #501, opened 2024-05-28
refactor!: Properly support contexts

bhavanki:
The new AWS SDK uses contexts for all API functions and for loading the
SDK configuration. When chamber was migrated to the SDK, all of them
were set to `context.TODO()`. This commit removes those temporary values
and threads contexts properly throughout the codebase, starting from the
chamber commands.

BREAKING CHANGE: This commit changes the signature of many exported
functions to add a context argument.

### DIFF
--- a/environ/environ.go
+++ b/environ/environ.go
@@ -1,6 +1,7 @@
 package environ
 
 import (
+	"context"
 	"fmt"
 	"strings"
 
@@ -83,8 +84,8 @@ func normalizeEnvVarName(k string) string {
 
 // load loads environment variables into e from s given a service
 // collisions will be populated with any keys that get overwritten
-func (e *Environ) load(s store.Store, service string, collisions *[]string) error {
-	rawSecrets, err := s.ListRaw(utils.NormalizeService(service))
+func (e *Environ) load(ctx context.Context, s store.Store, service string, collisions *[]string) error {
+	rawSecrets, err := s.ListRaw(ctx, utils.NormalizeService(service))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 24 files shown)


environ/environ.go
```diff
@@ -108,34 +109,34 @@ func (e *Environ) load(s store.Store, service string, collisions *[]string, noPa
 
 // Load loads environment variables into e from s given a service
 // collisions will be populated with any keys that get overwritten
-func (e *Environ) Load(s store.Store, service string, collisions *[]string) error {
-	return e.load(s, service, collisions, false)
+func (e *Environ) Load(ctx context.Context, s store.Store, service string, collisions *[]string) error {
```
alecjacobs5401, line 112: as it stands, anywhere we add in `ctx` here as a starting argument is considered a breaking change since it alters the public interface.

I know we're preparing the default branch as a candidate for v3, so maybe its moot, but does it make sense to call this out heavily somewhere (either in a changelog, or updating the merge commit) to designate the nature of these breaking changes?

store/s3store.go
```diff
@@ -58,18 +55,8 @@ type S3Store struct {
 	bucket string
 }
 
-// Deprecated; use NewS3StoreWithBucket instead
-func NewS3Store(numRetries int) (*S3Store, error) {
```
alecjacobs5401, line 62: this feels like a sneak in that isn't necessarily related to supporting contexts - does this deserve its own PR/changelog line item for the v3 release?

alecjacobs5401 (commented):
Changes all seem reasonable here! Mainly have a a recurring open question about the nature of this PR introducing breaking changes to almost every public interface. I know this is slated for a v3 release, but should this PR be split up/the title include the nature of the breaking changes? I believe the CHANGELOG is auto generated based on commits now, so want to make sure we're building up a good changelog for migrations. 

I know its more work too, but a well worn pattern I've seen is to add a `<Action>WithContext` as a way to side-step the breaking nature of passing contexts through. Is that something we might need to consider?

environ/environ.go
bhavanki: I'm fine updating the commit to be a breaking change in conventional commit-ese. It is certainly a significant break for anyone using chamber as a library, which I wasn't thinking about.

store/s3store.go
bhavanki: From the integrator standpoint, yes. Let me spin off the removal into a separate PR that we can merge before this one.

It's related because removing it avoids needing to add `ctx` to it. The constructor has been deprecated for over five years.

bhavanki: https://github.com/segmentio/chamber/pull/502

environ/environ.go
bhavanki: Commit updated to call out that this is a breaking change.

bhavanki: As for `<Action>WithContext`: I'm OK with using the breaking change opportunity to avoid adding those extra functions to preserve compatibility. It'd certainly be a lot of extra functions.

alecjacobs5401 (dismissed):

Sabrina0614 (approved):

alecjacobs5401 (approved):

PR #503, opened 2024-06-03
chore!: Remove deprecated support for non-path SSM API

bhavanki:
Version 2.0 of chamber deprecated use of the non-path-based API for the
SSM parameter store. This commit removes support for it completely.

Since the deprecation occurred so long ago, this changeset does not
preserve the ability to migrate to using the path-based API. Migration
may be performed with a 2.x version of chamber prior to upgrading to
a 3.x version.

BREAKING CHANGE: Any users or clients still using the non-path-based SSM API will see errors working with the SSM parameter store.

### DIFF
--- a/store/ssmstore_test.go
+++ b/store/ssmstore_test.go
@@ -262,9 +262,8 @@ func matchStringFilters(filters []types.ParameterStringFilter, param mockParamet
 	return true, nil
 }
 
-func NewTestSSMStore(parameters map[string]mockParameter, usePaths bool) *SSMStore {
+func NewTestSSMStore(parameters map[string]mockParameter) *SSMStore {
 	return &SSMStore{
-		usePaths: usePaths,
 		svc: &apiSSMMock{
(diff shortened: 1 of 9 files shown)


store/ssmstore_test.go
```diff
@@ -262,9 +262,8 @@ func matchStringFilters(filters []types.ParameterStringFilter, param mockParamet
 	return true, nil
 }
 
-func NewTestSSMStore(parameters map[string]mockParameter, usePaths bool) *SSMStore {
```
alecjacobs5401, line 265: I love how this test had a `usePaths` boolean argument but the actual code used a `noPaths` boolean argument. Twisted my brain right up!

alecjacobs5401 (approved):

store/ssmstore_test.go
bhavanki: Oh geez, I don't know if I even noticed that

PR #504, opened 2024-06-05
add note about secrets manager

calebAtIspot:
#267  added secret manager support, so I updated the readme to note that. That way people won't need to dig through the issues to figure out what the support status is.

I've checked "allow edits by maintainers", so feel free to change the wording / formatting :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,14 +412,6 @@ chamber -b secretsmanager write myservice foo fah
 chamber -b secretsmanager write myservice foo2 fah2
 ```
 
-This will result in one secret being generated with the following JSON value:
-
-```json
-{"_chamber_metadata":"{\"foo\":{\"created\":\"2024-06-05T00:32:06.680112Z\",\"created_by\":\"arn:aws:sts::0123456789:assumed-role/yourrole\",\"version\":1},\"foo2\":{\"created\":\"2024-06-05T00:32:39.672526Z\",\"created_by\":\"arn:aws:sts::0123456789:assumed-role/yourrole\",\"version\":1}}",
-"foo":"fah",
-"foo2":"fah2"}
-```
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -404,6 +404,22 @@ you can use `CHAMBER_AWS_REGION` to override just for chamber.
...
 
+## AWS Secrets Manager
+Chamber supports AWS Secrets Manager as an optional backend. For example:
+
+```
+chamber -b secretsmanager write myservice foo fah
+chamber -b secretsmanager write myservice foo2 fah2
+```
+
+This will result in one secret being generated with the following JSON value:
```
bhavanki, line 415: It'd be better for the README not to commit to / advertise the particular JSON format that the secrets manager backend uses. Based on comments in the original PR and the general sentiment I get from the code, there's no guarantee of interoperability in how chamber stores things, so we'd want to keep the freedom to revamp the format used for Secrets Manager in the future. (Plus we'd have to keep updating this doc!)

If dropping this makes the new section feel too small, it could be replaced with anything a chamber user needs to configure in Secrets Manager for chamber to work with it.

bhavanki (commented):
Thanks for addressing this documentation gap!

I'm assuming that @danmactough is still maintaining this backend, and that it is not experimental like the S3 backends. My assumptions could be wrong, but I feel optimistic today 😁

danmactough (approved):
> Thanks for addressing this documentation gap!
> 
> I'm assuming that @danmactough is still maintaining this backend, and that it is not experimental like the S3 backends. My assumptions could be wrong, but I feel optimistic today 😁

"maintaining" is a bit strong as these rare PRs about the secrets manager integration are the limit of my involvement 😄

bhavanki (approved):

PR #507, opened 2024-06-07
feat: Add tag read, write, and delete commands

bhavanki:
The new commands are for reading, writing, and deleting tags on the
underlying resources representing secrets.

For now, the commands are only implemented for SSM parameters.

### DIFF
--- a/cmd/tag-write.go
+++ b/cmd/tag-write.go
@@ -1,0 +1,95 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
+	analytics "github.com/segmentio/analytics-go/v3"
+	"github.com/segmentio/chamber/v2/store"
+	"github.com/segmentio/chamber/v2/utils"
+	"github.com/spf13/cobra"
+)
+
+var (
+	deleteOtherTags bool
+
+	// tagWriteCmd represents the tag read command
+	tagWriteCmd = &cobra.Command{
+		Use:   "write <service> <key> <tag>...",
+		Short: "Write tags for a specific secret",
+		Args:  cobra.MinimumNArgs(3),
+		RunE:  tagWrite,
+	}
+)
+
+func init() {
+	tagWriteCmd.Flags().BoolVar(&deleteOtherTags, "delete-other-tags", false, "Delete tags not specified in the command")
+	tagCmd.AddCommand(tagWriteCmd)
+}
+
+func tagWrite(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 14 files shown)


store/ssmstore.go
```diff
@@ -126,6 +127,72 @@ func (s *SSMStore) Read(ctx context.Context, id SecretId, version int) (Secret,
 	return s.readVersion(ctx, id, version)
 }
 
+func (s *SSMStore) WriteTags(ctx context.Context, id SecretId, tags map[string]string, deleteOtherTags bool) error {
+	// list the current tags
+	currentTags, err := s.ReadTags(ctx, id)
```
alecjacobs5401, line 132: we only use the current tags if we're doing an idempotent tag update and deleting other tags, right?

If so, can we only ready the tags when we're going to do something with those current tags?

cmd/tag-write.go
```diff
@@ -0,0 +1,95 @@
...
+	tagWriteCmd = &cobra.Command{
+		Use:   "write <service> <key> <tag>...",
+		Short: "Write tags for a specific secret",
+		Args:  cobra.MinimumNArgs(3),
+		RunE:  tagWrite,
+	}
+)
+
+func init() {
+	tagWriteCmd.Flags().BoolVar(&deleteOtherTags, "delete-other-tags", false, "Delete tags not specified in the command")
```
alecjacobs5401, line 28: This is a descriptive variable name, but I'm curious if there is precedent for CLIs to have an argument that is semantically the same but more conventional? Maybe `--reset` or `--sync` or `--prune`? Really not asking for a change just opening up a discussion about what a conventional flag someone might expect to exist on a CLI that has this behavior in a low surprise manner

alecjacobs5401 (commented):
Few open questions - nothing really jumps out at me immediately here otherwise

cmd/tag-write.go
bhavanki: I didn't really do a survey, but I did ask OpenAI for suggestions and chose this 😆

store/ssmstore.go
bhavanki: Good catch! Fixed.

cmd/tag-write.go
alecjacobs5401: Hah! Funny. I asked open AI the same thing and it suggested those (specifically sync and prune) 😂

alecjacobs5401 (approved):

cmd/tag-write.go
bhavanki: I guess it depends on the question phrasing and OpenAI's mood that day!

PR #516, opened 2024-06-10
feat: Add tag-on-create

bhavanki:
Store implementations may support tagging secrets immediately after they
are created, as a convenience and a basis for future enforcement of
tagging. As usual, only the SSM store implementation supports the new
feature.

The README is updated with documentation about both this new feature and
the recently added tag commands.

### DIFF
--- a/store/ssmstore.go
+++ b/store/ssmstore.go
@@ -89,6 +89,14 @@ func (s *SSMStore) KMSKey() string {
 // Write writes a given value to a secret identified by id.  If the secret
 // already exists, then write a new version.
 func (s *SSMStore) Write(ctx context.Context, id SecretId, value string) error {
+	return s.write(ctx, id, value, nil)
+}
+
+func (s *SSMStore) WriteWithTags(ctx context.Context, id SecretId, value string, tags map[string]string) error {
+	return s.write(ctx, id, value, tags)
+}
+
+func (s *SSMStore) write(ctx context.Context, id SecretId, value string, tags map[string]string) error {
 	version := 1
 	// first read to get the current version
 	current, err := s.Read(ctx, id, -1)
@@ -97,6 +105,10 @@ func (s *SSMStore) Write(ctx context.Context, id SecretId, value string) error {
 	}
 	if err == nil {
 		version = current.Meta.Version + 1
+	}
+
+	if len(tags) > 0 && version != 1 {
+		return errors.New("tags on write only supported for new secrets")
 	}
 
 	putParameterInput := &ssm.PutParameterInput{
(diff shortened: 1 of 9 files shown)


store/ssmstore.go
```diff
@@ -99,6 +107,10 @@ func (s *SSMStore) Write(ctx context.Context, id SecretId, value string) error {
 		version = current.Meta.Version + 1
 	}
 
+	if len(tags) > 0 && version != 1 {
+		return errors.New("tags on write only supported for new secrets")
+	}
```
Sabrina0614, line 112: ❓ Curious if there is a particular reason why tagging is only support for new secrets? Sorry if I am missing it somewhere

store/ssmstore.go
```diff
@@ -114,6 +126,12 @@ func (s *SSMStore) Write(ctx context.Context, id SecretId, value string) error {
 		return err
 	}
 
+	if len(tags) > 0 {
+		if err := s.WriteTags(ctx, id, tags, false); err != nil {
```
Sabrina0614, line 130: ❓ (More of the WriteTags method) - It looks like the WriteTags also take in `deleteOtherTags`, curious what would happen for the following case:
1. If there already exist a tag for a secret, let's say it has tag`app = backstage` and another tag `requiredChamber = true`
2. I then now call write tag for this secret with the `app = backstage`  and with `deleteOtherTags` is set to true, it will remove the `required = chamber`, but would it error out if we are trying to add the tag that already exist?

Sabrina0614 (commented):
Looks great! Just some questions. Thank you for adding it and also update the README

store/ssmstore.go
bhavanki: You didn't miss anything! (I thought for a second that I had documented why, but it's just in my notes.) I was thinking of what chamber should do when the secret already exists and it already has tags.

* If you include `--tags`, what should happen to the other tags that you don't list? Should they be deleted or left alone? Admittedly, we could add the `--delete-other-tags` option from `chamber tag write` to let you pick.
* If you do not include `--tags`, what should happen to the existing tags? Should they be all dropped out or left alone? We could maybe add another option, `--delete-existing-tags` or something, to let you pick.

It's because of this complexity that, at least for this pass, I opted to restrict tags to just new secrets.

store/ssmstore.go
bhavanki: If I understand your scenario correctly, it would be fine, because you can use `chamber tag write` to replace the value of an existing tag, and it's fine if the new value is the same as the old value.

In the implementation of `WriteTags`, first all of the tags which are not being written are deleted; because the `app` tag is being written, it isn't deleted. Then, the `app` tag is updated using the SSM add-tags-to-resource API, which either adds or overwrites tags.

Let me know if I misunderstood your scenario!

Sabrina0614: makes sense! thank you for walking me through it!

store/ssmstore.go
Sabrina0614: thank you for the context! I did not know that it would override the existing tag prior.

Sabrina0614 (approved):

PR #517, opened 2024-06-10
refactor: Switch to AWS endpoint resolution v2

bhavanki:
The v2 AWS SDK introduces a new endpoint resolution mechanism where
custom endpoints are specified on a per-service basis, while the
previous global mechanism is deprecated. This changeset switches to the
newer mechanism.

The default resolver under the new mechanism uses the base endpoint in
a service client's `Options` as the means to customize endpoints, so
it's enough for tests to make sure the base endpoint gets set.

Under the previous mechanism, the CHAMBER_AWS_SSM_ENDPOINT environment
variable was used to customize not just the endpoint for SSM, but also
for Secrets Manager. This changeset deprecates the latter use in favor
of a new CHAMBER_AWS_SECRETS_MANAGER_ENDPOINT environment variable.
There seems to also be SDK-native ways to use the environment for
endpoint customization, so this whole thing might go away at some point.

### DIFF
--- a/store/secretsmanagerstore.go
+++ b/store/secretsmanagerstore.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"os"
 	"reflect"
 	"sort"
 	"strconv"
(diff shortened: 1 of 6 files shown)


Sabrina0614 (approved):

bhavanki:
I inadvertently left the old resolver in place in shared.go. Fixing!

Sabrina0614 (approved):

PR #519, opened 2024-06-17
chore: Convert to log/slog

bhavanki:
The small amount of verbose logging in the exec command has been
converted to debug logging under log/slog.

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"errors"
 	"fmt"
+	"log/slog"
 	"os"
 	"strings"
 
@@ -97,15 +98,13 @@ func execRun(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Failed to get secret store: %w", err)
 	}
 
-	if pristine && verbose {
-		fmt.Fprintf(os.Stderr, "chamber: pristine mode engaged\n")
+	if pristine {
+		slog.Debug("chamber: pristine mode engaged")
 	}
 
 	var env environ.Environ
 	if strict {
-		if verbose {
-			fmt.Fprintf(os.Stderr, "chamber: strict mode engaged\n")
-		}
+		slog.Debug("chamber: strict mode engaged")
 		var err error
 		env = environ.Environ(os.Environ())
 		err = env.LoadStrict(cmd.Context(), secretStore, strictValue, pristine, services...)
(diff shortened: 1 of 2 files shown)


bhavanki:
slog was introduced in Go 1.21, so this cannot be integrated yet since chamber still supports go 1.20. I'll leave this PR inactive until the 1.23 release, which is expected in August (about 2 months from now).

Sabrina0614 (approved):

cmd/root.go
```diff
@@ -72,19 +75,22 @@ var RootCmd = &cobra.Command{
...
 	RootCmd.PersistentFlags().DurationVarP(&minThrottleDelay, "min-throttle-delay", "", 0, "DEPRECATED and no longer has any effect. Use retry-mode instead")
+	RootCmd.PersistentFlags().MarkDeprecated("min-throttle-delay", "use --retry-mode instead")
 	RootCmd.PersistentFlags().StringVarP(&retryMode, "retry-mode", "", store.DefaultRetryMode.String(),
 		`For SSM, the model used to retry requests
   `+aws.RetryModeStandard.String()+`
   `+aws.RetryModeAdaptive.String(),
 	)
 	RootCmd.PersistentFlags().BoolVarP(&verbose, "verbose", "", false, "Print more information to STDOUT")
+	RootCmd.PersistentFlags().MarkDeprecated("verbose", "use --log-level debug instead")
+	RootCmd.PersistentFlags().StringVar(&logLevel, "log-level", "", "Log level")
```
alecjacobs5401, line 93: as a user of chamber, how do I know what the valid values for log-level is? Do we _need_ this level of granularity versus just a verbose flag?

bhavanki: Removed `--log-level` and un-deprecated `--verbose`.

cmd/exec.go
```diff
@@ -130,9 +129,7 @@ func execRun(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	if verbose {
-		fmt.Fprintf(os.Stdout, "info: With environment %s\n", strings.Join(env, ","))
-	}
+	slog.Debug(fmt.Sprintf("info: With environment %s\n", strings.Join(env, ",")))
```
alecjacobs5401, line 132: is there any world where this is considered a breaking change? It looks like the comment in the help output for verbose before was wrong since partially we output to STDERR (only this line apparently was to STDOUT)

Do we need to keep that consistency for any reason?

alecjacobs5401 (approved):
moreso an open question, but thanks for doing this!

kevinburkesegment:
@bhavanki Let's merge this

PR #532, opened 2024-07-03
feat!: Add store configuration support for SSM store

bhavanki:
The new Config and SetConfig methods on the Store interface allow
implementations to maintain their own configurations. Only the SSM store
fully implements the methods; the others return an empty configuration
and do not support setting one.

The store configuration itself holds a list of required tags for each
written secret. Enforcement of the tags is not yet implemented. Although
the configuration struct is exported from its package, it is not part of
chamber's client interface and is subject to change at any time.

The SSM store implementation stores its configuration as a secret
inside the newly reserved "_chamber" service as a JSON document. The
schema is not exported, so users shouldn't build anything from it.

### DIFF
--- a/store/nullstore.go
+++ b/store/nullstore.go
@@ -13,6 +13,16 @@ func NewNullStore() *NullStore {
 	return &NullStore{}
 }
 
+func (s *NullStore) Config(ctx context.Context) (StoreConfig, error) {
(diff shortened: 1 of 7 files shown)


sindhusegment (approved):
thanks for explaining this in stand up!

PR #633, opened 2025-08-07
replaces broken aws blog link with segment.com

### LINKED ISSUE #597: Link in README to AWS Parameter Store document is broken
Apologies if this is already known, I was reading the REAMDE and noticed that the link for `The Right Way To Manage Secrets` is broken, and I couldn't easily find what the updated link _should_ be on AWS's documentation maze. Wanted to make sure that this was known in case anyone else tries to use that link and runs into the same thing.

jtzero:
fixes https://github.com/segmentio/chamber/issues/597 (Link in README to AWS Parameter Store document is broken)


eldondevat:
This link is now also broken!

PR #646, opened 2025-10-02
Add --no-warn-conflicts flag

mlanett:
Let us improve the use cases to support inheritance of secrets. For example:
- defaults/myservice/…
- staging/myservice/…
- production/myservice/…

At startup time, a service loads its secrets using chamber with multiple paths, e.g. `chamber exec defaults/myservice staging/myservice -- …`. This leads to many warnings of overwritten variables. However we are intentionally overwriting them, so the warnings do not reveal problems but instead pollute service logs. Thus we propose a flag --no-warn-conflicts to suppress these warnings.


TimSpence (approved):
👍

TimSpence:
Can we merge this contribution from @mlanett ?

artemshakhovpj (approved):

javierortizpj:
Bump - This would be helpful in cleaning up our logs.

sushma-rayasam (approved):

